PR #75, opened 2021-09-26
Hatpooltokenwrapper POC

jellegerbrandy:


contracts/HATVaultsTokenWrapper.sol
```diff
@@ -0,0 +1,46 @@
...
+    require(msg.sender == poolContract);
+    wrappedToken.safeTransfer(recipient, amount);
+    return true;
+  }
+
+  function approve(address spender, uint256 amount) public returns (bool)  {
+    require(msg.sender == poolContract);
+    return wrappedToken.approve(spender, amount);
+  }
+   function balanceOf(address account) public view returns (uint256) {
```
orenyodfat, line 42: why need balanceOf ?

contracts/HATVaultsTokenWrapper.sol
```diff
@@ -0,0 +1,46 @@
...
+    address public immutable poolContract;
+     constructor(address _tokenToWrap, address _poolContract) {
+        wrappedToken = IERC20(_tokenToWrap);
+        poolContract = _poolContract;
+
+     }
+
+  function transferFrom(address _sender, address _recipient, uint256 _amount) public returns (bool) {
+    require(msg.sender == poolContract, "only poolContract can call transferFrom");
+    require(_recipient != address(this));
```
orenyodfat, line 23: is this needed ?

contracts/HATVaultsTokenWrapper.sol
```diff
@@ -0,0 +1,46 @@
...
+
+    address recipient = _recipient;
+    if (_recipient == poolContract) {
+      recipient = address(this);
```
orenyodfat, line 27: no condition needed as recipient will be always poolContract

contracts/HATVaultsTokenWrapper.sol
```diff
@@ -0,0 +1,46 @@
...
+import "openzeppelin-solidity/contracts/token/ERC20/utils/SafeERC20.sol";
+
+
+ // Wrap an ERC20 token with other ERC20 token.
+contract HATVaultsTokenWrapper {
+    using SafeERC20 for IERC20;
+
+    IERC20 public immutable wrappedToken;
+    address public immutable poolContract;
+     constructor(address _tokenToWrap, address _poolContract) {
```
orenyodfat, line 15: lets add here token symbol,name and decimal .. so it will be manifested in the UI.

contracts/HATVaultsTokenWrapper.sol
```diff
@@ -0,0 +1,46 @@
...
+        wrappedToken = IERC20(_tokenToWrap);
+        poolContract = _poolContract;
+
+     }
+
+  function transferFrom(address _sender, address _recipient, uint256 _amount) public returns (bool) {
+    require(msg.sender == poolContract, "only poolContract can call transferFrom");
```
orenyodfat, line 22: also uniswap v3 should be able to transferFrom https://github.com/Uniswap/v3-core/blob/main/contracts/UniswapV3Pool.sol#L773 . this is the reason it is approve at https://github.com/hats-finance/hats-contracts/blob/develop/contracts/HATVaults.sol#L755

contracts/HATVaultsTokenWrapper.sol
orenyodfat: oh. need for tokenLock

contracts/HATVaultsTokenWrapper.sol
```diff
@@ -0,0 +1,46 @@
...
+
+    address recipient = _recipient;
+    if (_recipient == poolContract) {
+      recipient = address(this);
+    }
+    wrappedToken.safeTransferFrom(_sender, recipient, _amount);
+    return true;
+  }
+  function transfer(address recipient, uint256 amount) public returns (bool) {
+    require(msg.sender == poolContract);
```
orenyodfat, line 33: transfer can be called also from tokenLock contract

contracts/HATVaultsTokenWrapper.sol
```diff
@@ -0,0 +1,65 @@
...
+  function transferFrom(address _sender, address _recipient, uint256 _amount) public override returns (bool) {
+    // if the recipient is the poolContract, we create new wrapped tokens, and add the sender to the list of stakers
+    // if the recipient is a staker, we unwrap the tokens and send them to the staker
+    // in all other cases, we just use the ERC20 transferFrom
+    if (_recipient == poolContract) {
+      wrappedToken.safeTransferFrom(_sender, address(this), _amount);
+      _mint(_recipient, _amount);
+      stakers[_sender] = true;
+    } else {
+      // if (stakers[_recipient]) {
```
orenyodfat, line 39: is that comment out in purpose ?

contracts/HATVaultsTokenWrapper.sol
```diff
@@ -0,0 +1,65 @@
...
+  constructor(address _tokenToWrap, address _poolContract, string memory name_, string memory symbol_) ERC20(name_, symbol_) {
+    wrappedToken = IERC20(_tokenToWrap);
+    poolContract = _poolContract;
+  }
+
+  function transferFrom(address _sender, address _recipient, uint256 _amount) public override returns (bool) {
+    // if the recipient is the poolContract, we create new wrapped tokens, and add the sender to the list of stakers
+    // if the recipient is a staker, we unwrap the tokens and send them to the staker
+    // in all other cases, we just use the ERC20 transferFrom
+    if (_recipient == poolContract) {
```
orenyodfat, line 34: and if the safeTransferFrom come from rewardDepositors https://github.com/hats-finance/hats-contracts/blob/develop/contracts/HATVaults.sol#L293 ?  
it is a potential attack vector

contracts/HATVaultsTokenWrapper.sol
```diff
@@ -0,0 +1,65 @@
...
+      // }
+    }
+    return true;
+  }
+  function transfer(address _recipient, uint256 _amount) public override returns (bool) {
+    // if recipient is a staker, we unwrap the tokens 
+    // in other case, we just call the usual transfer
+    if (stakers[_recipient]) {
+      _burn(msg.sender, _amount);
+      wrappedToken.safeTransfer(_recipient, _amount);
```
orenyodfat, line 53: so an attacker Staker can deposit 1 token , then can transfer to her self all other victims deposits.

contracts/HATVaultsTokenWrapper.sol
```diff
@@ -0,0 +1,65 @@
...
+      // } else {
+        super.transferFrom(_sender, _recipient, _amount);
+      // }
+    }
+    return true;
+  }
+  function transfer(address _recipient, uint256 _amount) public override returns (bool) {
+    // if recipient is a staker, we unwrap the tokens 
+    // in other case, we just call the usual transfer
+    if (stakers[_recipient]) {
```
orenyodfat, line 51: so any staker can call transfer and withdraw its deposit bypassing hatsVault?

jellegerbrandy: the next line says that msg.sender  burns _amount of wrapper tokens. A staker typically has 0 wrapper tokens - see lin 36, which is the only point in which wrapper tokens are minted. So no I do not think you are right..

contracts/HATVaultsTokenWrapper.sol
jellegerbrandy: yes, so I was not sure this was the "expected behavior" and it did not seem to matter for the tests

contracts/HATVaultsTokenWrapper.sol
jellegerbrandy: I do not see how. So that would mean that the caller of the rewardDepositors function is put on the stakers list while he is not a staker. Which is not an attack vector, afaics (mb I should rename the from stakers to `accountsThatKeepTheirTokensUnwrapped`

PR #79, opened 2021-12-09
Add Hat timelock controller

ben-kaufman:
Add timelock controller for hat vaults

### DIFF
--- a/contracts/HATTimelockController.sol
+++ b/contracts/HATTimelockController.sol
@@ -12,22 +12,20 @@ contract HATTimelockController is TimelockController {
     HATVaults public hatVaults;
 
     constructor(
+        HATVaults _hatVaults,
         uint256 _minDelay,
         address[] memory _proposers,
         address[] memory _executors
-    // solhint-disable-next-line func-visibility, no-empty-blocks
+    // solhint-disable-next-line func-visibility
     ) TimelockController(_minDelay, _proposers, _executors) {
-    }
-
-    function initializeHATVaults(HATVaults _hatVaults) external onlyRole(EXECUTOR_ROLE) {
         require(address(_hatVaults) != address(0), "HATTimelockController: HATVaults address must not be 0");
-        require(address(hatVaults) == address(0), "HATTimelockController: Cannot update HATVaults address");
         hatVaults = _hatVaults;
+
     }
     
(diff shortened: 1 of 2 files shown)


contracts/HATTimelockController.sol
```diff
@@ -0,0 +1,82 @@
...
+
+    constructor(
+        uint256 _minDelay,
+        address[] memory _proposers,
+        address[] memory _executors
+    // solhint-disable-next-line func-visibility, no-empty-blocks
+    ) TimelockController(_minDelay, _proposers, _executors) {
+    }
+
+    function initializeHATVaults(HATVaults _hatVaults) external onlyRole(EXECUTOR_ROLE) {
```
jellegerbrandy, line 22: why executor role? 
So I am not sure how this contract will be deployed, but the current suggestion in https://docs.google.com/document/d/1raskPjnNeHQMaOfAgCIzgv7NTTH5uRd2ZBj0yMhSMWM/edit# is to give anyone the executor role, which would not be good. 
Also: is it necessary to be able to reset this? I guess that if we upgrade the hatvaults contract, you would also deploy a new timelockcontroller. Mb just remove this function

contracts/HATTimelockController.sol
```diff
@@ -0,0 +1,82 @@
...
+        require(address(_hatVaults) != address(0), "HATTimelockController: HATVaults address must not be 0");
+        require(address(hatVaults) == address(0), "HATTimelockController: Cannot update HATVaults address");
+        hatVaults = _hatVaults;
+    }
+    
+    // Whitelisted functions
```
jellegerbrandy, line 28: I believe it makes more sense sot give all whitelisted functions the PROPOSER_ROLE, but let's discuss as it can also go this way

jellegerbrandy: image

shayzluf (commented):
@ben-kaufman could you please fix github actions errors?
https://github.com/hats-finance/hats-contracts/pull/79/commits/c71448fd198e5341737ebd939504777dd50c742f

jellegerbrandy (approved):
ok looks good to me, very straightforward in this form

maybe a few lines of documentation would be good, to explain what the ocntrat does

PR #83, opened 2022-01-17
estimate gas costs

shayzluf:
Add gas estimation calculation to tests


.gitignore
```diff
@@ -14,6 +14,7 @@ coverage.json
 cache/
 yarn*
 frontend/
+.env
```
jellegerbrandy, line 17: +1 for using dotenv, we should remove the .private.json file though as we now have 2 files with configs

jellegerbrandy (approved):
looks good to me but it is a bit a jump in the dark as the deploy script is not tested..

PR #84, opened 2022-01-17
Timelock configuration and other tweaks [wip]

jellegerbrandy:


README.md
```diff
@@ -10,25 +8,30 @@ The Hats protocol is designed to give white hats hackers the opportunity to gain
...
 
- Installation:
+```
+npm install
+```
 
- ```
- npm install
- ```
+Create `.env` and `.private.json` files as needed.
```
ben-kaufman, line 17: This is duplicate and very confusing, better have just one (probably .env) and remove the other file

scripts/addresses.js
```diff
@@ -0,0 +1,22 @@
...
+    oldTimelock: "0x66922e992e07030CEaC25E1919E9C31153F85b6f",
+    timelock: "0xFd4255F16378306CA83E37015Df01a1700DAc296",
+    newTimelock: "0xFd4255F16378306CA83E37015Df01a1700DAc296",
+  },
+  rinkeby: {
+    governance: "0xd6eC2F1D35C4a21D3a25E95CC40D67b4D0af995E",
+    hatVaultsAddress: "0x59D4316A4691ceB657347dc7b0849e0CB38ed106",
+    timelock: "0xa0Af770709377D0c6cfFc72BeD965A6674244C76", // this is patched timelock
+    executors: ["0x2bc1fed4c65c9b1dc2baaff2f3198acc42c41778"],
+  },
```
ben-kaufman, line 21: Should add hardhat config for testing locally the script

scripts/timelock-deploy.js
```diff
@@ -0,0 +1,118 @@
...
+  // the deployer automatically gets the admin role, which we renounce (so all admin has to go through the proposer)
+  await renounceRole({
+    timelock: hatTimelockController.address,
+    deployer: deployer.address,
+  });
+
+  // We also save the contract's artifacts and address in the frontend directory
+  if (network.name !== "hardhat") {
+    await verifyTimelock(
+      config.update({ timelock: hatTimeLockController.address })
```
ben-kaufman, line 69: This update method doesn't seem to exist. Should update like: `config["timelock"] = hatTimelockController.address;` on the line before then just pass config.

jellegerbrandy: what version of node are u using?

README.md
jellegerbrandy: yes

scripts/addresses.js
jellegerbrandy: Ok so I did, but these scripts are not very generic - for example, deployvaults has all kind of values for rinkeby hardcoded in. To fix that I think is a separate project to do in a separate pr.

scripts/timelock-deploy.js
jellegerbrandy: anway this script is tested, so if CI says it is ok then the problem is yours :)

ben-kaufman: Oh yeah bad node version sorry

ben-kaufman (approved):

ben-kaufman (approved):

PR #85, opened 2022-02-09
Hats v1.1 [WIP]

ben-kaufman:
- Add restrictions on certain parameters values  (addresses #87 and #89)
- Allow governance to set the reward multipliers  (#92)
- Begin to optimize require messages to reduce contract size ((#91)


contracts/HATVaults.sol
```diff
@@ -9,7 +9,40 @@ import "./HATMaster.sol";
 import "./tokenlock/ITokenLockFactory.sol";
 import "./Governable.sol";
 
-
+// Errors:
+// HVE01: Only committee
```
jellegerbrandy, line 13: are there no tests for there error messages?

test/hatvaults.js
```diff
@@ -301,6 +301,36 @@ contract("HatVaults", (accounts) => {
...
+    });
+    try {
+      await hatVaults.dismissPendingApprovalClaim(0, { from: accounts[1] });
+      assert(false, "only governance can dismiss before delay");
+    } catch (ex) {
+      assertVMException(ex);
+    }
+    await utils.increaseTime(1);
+    await utils.increaseTime(5 * 7 * 24 * 60 * 60);
+    await hatVaults.dismissPendingApprovalClaim(0, { from: accounts[1] });
```
jellegerbrandy, line 331: please also test that this actually dismisses the claim

jellegerbrandy (changes requested):
- So I thinkn that given that there are still open questions regarding #92 , it would be better to separate those changes out into a new PR (this is better so that the resoltion of the other issues i not blocked)
- Definitely, the changes regarding the 1 token per pool thing #90 should be in a separate PR.

jellegerbrandy (changes requested):
Ok just 2 small things still

jellegerbrandy (approved):

PR #97, opened 2022-03-03
Disallow changing reward multipliers las value

### LINKED ISSUE #92: Allow governance to set rewardmultipliers
the rewardmultipliers are currently hardcoded, they need to be settable by governance

ben-kaufman:
Fix #92

### DIFF
--- a/contracts/HATVaults.sol
+++ b/contracts/HATVaults.sol
@@ -340,7 +340,7 @@ contract  HATVaults is Governable, HATMaster {
      * @dev setRewardMultipliers - called by hats governance to set reward multipliers
      * @param _rewardMultipliers reward multipliers
     */
-    function setRewardMultipliers(uint256[25] memory _rewardMultipliers) external onlyGovernance {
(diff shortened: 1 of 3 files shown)


contracts/HATVaults.sol
```diff
@@ -340,7 +340,7 @@ contract  HATVaults is Governable, HATMaster {
      * @dev setRewardMultipliers - called by hats governance to set reward multipliers
      * @param _rewardMultipliers reward multipliers
     */
-    function setRewardMultipliers(uint256[25] memory _rewardMultipliers) external onlyGovernance {
+    function setRewardMultipliers(uint256[24] memory _rewardMultipliers) external onlyGovernance {
```
jellegerbrandy, line 343: Ok this is smart and will work, but it a bit implicit. I think it is a better pattern to use to make `rewardMultipliers` an array of 24 items, and change the logic in `getMultiplier`, wdyt?

ben-kaufman: Yes that'd be the full solution I guess. Changed it now.

jellegerbrandy (approved):

PR #98, opened 2022-03-09
Allow multiple pools with the same token

### LINKED ISSUE #90: Remove the 1 token 1 vault limit
Currently, each pool uses a unique token (i.e. it is not possible to create two pools that use the same token). 

This limit is easily removable by removing the `poolId1` mapping from the code.

However, we must also change the logic in the `swapAndBurn` function - if we leave that unchanged, and there are two different outstanding claims in the same token, the combined value of the two climes will go the beneficiary of the first (if I make myself clear). The solution here is to change the `swapAndBurns` to a mapping from  poolIds to amounts (rather than token addresses to amounts, as it is currently).  

We must carefully check if there are other places in the code that are affected. In the current version, the tokens in the pools are completely separated, which is a guarantee (or at least makes it hard otherwise) that paouts of one pool to not interfere with the balance of another pool.

ben-kaufman:
Fix #90

### DIFF
--- a/contracts/HATMaster.sol
+++ b/contracts/HATMaster.sol
@@ -16,7 +16,6 @@ import "openzeppelin-solidity/contracts/security/ReentrancyGuard.sol";
 // HME03: Committee not checked in yet
 // HME04: Withdraw: not enough user balance
 // HME05: User amount must be greater than 0
-// HME06: lpToken is already in pool
 contract HATMaster is ReentrancyGuard {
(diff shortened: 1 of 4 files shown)


jellegerbrandy (approved):
looks good to me. 
Just cannot merge bc test coverage diminished with 0.01% :)

jellegerbrandy (approved):

PR #112, opened 2022-03-23
Reset approval of token after

### LINKED ISSUE #100: Approval logic may break for USDT (and similar tokens)
USDT `approve` has a nonstandard implemention - it requires resetting approvals to 0 before setting it to a new value

See here: https://gist.github.com/plutoegg/a8794a24dfa84d0b0104141612b52977#file-tethertoken-sol-L265 (could not find a better reference, not sure why this contract is not on etherscan or I cannot find it)

This could be a problem for our `swapTokenForHAT`, on this line:

`require(_token.approve(address(uniSwapRouter), _amount), "HVE31");`

The solution is to either check, or set, the approval back to 0 after the token ahs been swapped away:

`require(_token.approve(address(uniSwapRouter),0), "..");`

ben-kaufman:
Fix #100

### DIFF
--- a/contracts/HATVaults.sol
+++ b/contracts/HATVaults.sol
@@ -44,6 +44,7 @@ import "./Governable.sol";
 // HVE32: Wrong amount received
 // HVE33: Reward level can not be more than 10000
 // HVE34: LP token is zero
+// HVE35: Token approve reset failed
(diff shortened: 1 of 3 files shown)


jellegerbrandy (approved):
lgtm

PR #113, opened 2022-03-24
Add withdrawal fee

### LINKED ISSUE #101: Fee on withdrawals
- user pays a fee (a percentage of his share) when withdrawing their stake
- fee is set per vault 
- fee is set as a percentage of the user's share (`function setFee(uint256 _pid, uint256 _newValue)`) and is payable in the vault token
- `setFee` is callable only by one address (`address feeSetter`) that is configurable by governance (in the timelock, I guess)
- feeSetter at first will be governance, but can be later set to another address

So we have:
- contract var `address feeSetter`
- function `setFeeSetter(address _newFeeSetter) onlyGovernance`
- modifier `onlyFeeSetter` which requires that the caller is either the `feeSetter`, or, if `feeSetter == address(0)`, the owner
- new function `function setFee(uint256 _pid, uint256 _newValue) onlyFeeSetter`,
  - fee is required to be < 2% (please doublecheck this)
- A change in `_withdraw`: fees are being sent to governance on withdrawal

ben-kaufman:
Fix #101

For now, the default fee for a pool is 0, and it does not apply for emergency withdrawal (specs are unclear about those).

### DIFF
--- a/contracts/HATVaults.sol
+++ b/contracts/HATVaults.sol
@@ -46,7 +46,7 @@ import "./Governable.sol";
 // HVE34: LP token is zero
 // HVE35: Token approve reset failed
 // HVE36: Only fee setter
-// HVE37: Fee must be less than 100%
+// HVE37: Fee must be less than 2%
 contract  HATVaults is Governable, HATMaster {
     using SafeMath  for uint256;
     using SafeERC20 for IERC20;
@@ -598,7 +598,7 @@ contract  HATVaults is Governable, HATMaster {
     }
 
     function setPoolFee(uint256 _pid, uint256 _newFee) external onlyFeeSetter {
-        require(_newFee < HUNDRED_PERCENT, "HVE36");
+        require(_newFee < MAX_FEE, "HVE36");
         poolInfo[_pid].fee = _newFee;
         emit SetPoolFee(_pid, _newFee);
     }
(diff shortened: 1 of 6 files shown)


contracts/HATVaults.sol
```diff
@@ -580,6 +591,17 @@ contract  HATVaults is Governable, HATMaster {
         emit SetPool(_pid, _allocPoint, _registered, _descriptionHash);
     }
 
+    function setFeeSetter(address _newFeeSetter) external onlyGovernance {
+        feeSetter = _newFeeSetter;
+        emit SetFeeSetter(_newFeeSetter);
+    }
+
+    function setPoolFee(uint256 _pid, uint256 _newFee) external onlyFeeSetter {
+        require(_newFee < MAX_FEE, "HVE36");
```
jellegerbrandy, line 600: lets do `<=`

jellegerbrandy (changes requested):

contracts/HATVaults.sol
```diff
@@ -44,7 +43,9 @@ import "./Governable.sol";
 // HVE32: Wrong amount received
 // HVE33: Reward level can not be more than 10000
 // HVE34: LP token is zero
-// HVE35: Token approve reset failed
+// HVE35: Only fee setter
+// HVE36: Fee must be less than 2%
+// HVE37: Token approve reset failed
 contract  HATVaults is Governable, HATMaster {
```
jellegerbrandy, line 49: Ah this still works even if you do not import it? Anyay, I'd remove the `Governanable` dependency here for clarity

jellegerbrandy (changes requested):

jellegerbrandy:
Ah, also the fee should apply for the emergency withdrawal as well, I guess

jellegerbrandy (approved):

PR #115, opened 2022-03-28
Remove minting of reward

### LINKED ISSUE #94: Deployment to Arbitrum
To make the system useable without paying too much gas, the vaults contracxt will be deployed to Arbitrum.  

Currently, the HatVaults contract will mint new tokens for rewarding vault share holders. 

We cannot, of course, mint new tokens on arbitrum. Instead, we will need to bridge Hats tokens over to abitrum and use those. 

The approach here, I guess, is to send hats tokens to the HatVaults contract, and change the the line at https://github.com/hats-finance/hats-contracts/blob/develop/contracts/HATMaster.sol#L143 and instead of minting there, transfer the tokens. We can consider implement on mainnet as well.

(another approach is to change the Arbitrum Hats contract, and implement the mint function as a transfer function. This is a bit riskier, and probably not acceptable as the token will have a different address as well)

The migration path would then be:

1. Deploy Hats erc20 contract to arbitrum
2. Change the HatVaults.sol contract, to call `transfer` instead of `mint` on https://github.com/hats-finance/hats-contracts/blob/develop/contracts/HATMaster.sol#L143.  (decide if we want to change that on mainnet as well or have two versions, or make it depend on the chainid)
3. Deploy hatvaults do arbitrum
4. Send HATs tokens over to the new vault

ben-kaufman:
Fix #94

### DIFF
--- a/contracts/HATMaster.sol
+++ b/contracts/HATMaster.sol
@@ -16,11 +16,9 @@ import "openzeppelin-solidity/contracts/security/ReentrancyGuard.sol";
 // HME03: Committee not checked in yet
 // HME04: Withdraw: not enough user balance
 // HME05: User amount must be greater than 0
-// HME06: HAT token cannot be used as the LP token
 contract HATMaster is ReentrancyGuard {
     using SafeMath for uint256;
     using SafeERC20 for IERC20;
-    using SafeERC20 for HATToken;
 
     struct UserInfo {
         uint256 amount;     // The user share of the pool based on the amount of lpToken the user has provided.
@@ -122,7 +120,7 @@ contract HATMaster is ReentrancyGuard {
 
     function depositHATReward(uint256 _amount) external {
         rewardAvailable += _amount;
-        HAT.safeTransferFrom(address(msg.sender), address(this), _amount);
+        HAT.transferFrom(address(msg.sender), address(this), _amount);
     }
     
   /**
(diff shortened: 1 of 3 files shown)


contracts/HATMaster.sol
```diff
@@ -16,9 +16,11 @@ import "openzeppelin-solidity/contracts/security/ReentrancyGuard.sol";
 // HME03: Committee not checked in yet
 // HME04: Withdraw: not enough user balance
 // HME05: User amount must be greater than 0
+// HME06: HAT token cannot be used as the LP token
```
jellegerbrandy, line 19: I guess that will be removed :)

contracts/HATMaster.sol
```diff
@@ -16,9 +16,11 @@ import "openzeppelin-solidity/contracts/security/ReentrancyGuard.sol";
...
 contract HATMaster is ReentrancyGuard {
     using SafeMath for uint256;
     using SafeERC20 for IERC20;
+    using SafeERC20 for HATToken;
```
jellegerbrandy, line 23: we do not need that I think, as we know HAtToken is "safe"

contracts/HATMaster.sol
```diff
@@ -323,11 +327,13 @@ contract HATMaster is ReentrancyGuard {
 
     // Safe HAT transfer function, just in case if rounding error causes pool to not have enough HATs.
     function safeTransferReward(address _to, uint256 _amount, uint256 _pid) internal {
-        uint256 hatBalance = HAT.balanceOf(address(this));
-        if (_amount > hatBalance) {
-            HAT.transfer(_to, hatBalance);
-            emit SendReward(_to, _pid, hatBalance, _amount);
+        uint256 _rewardAvailable = rewardAvailable;
```
jellegerbrandy, line 330: why copy the value?

contracts/HATMaster.sol
```diff
@@ -323,11 +327,13 @@ contract HATMaster is ReentrancyGuard {
...
+        if (_amount > _rewardAvailable) {
```
jellegerbrandy, line 331: what about doing a small refactor here and do:
```
if (_amount > rewardAvailable) { _amount = rewardAvailable }
rewardAvailable = rewardAvailable - _amount;
HAT.transfer(_to, _amount);
emit SendReward(_to, _pid, _amount, _amount);

jellegerbrandy (approved):

PR #136, opened 2022-05-03
Add multiple routers support

### LINKED ISSUE #99: Make path for selling vault tokens configurable
DXdao has raised an issue that regards the selling of DXD tokens from the vault for HATs in case a bounty is paid out. 

Currently, the swap is routed through the uniswap, using the ETH pair as an intermediate pair (so the swap is from VaultToken --> WETH --> HATs)

DXdao does not have much DXD/ETH liqudity on uniswapv3 (currently less than 6K$), and so this trade does not make much sense - even if there is enough liquidity to pay out a bug bounty, the amount lost to slippage would be prohibitive.

The proposal is to make the routing configurable by the caller 
The action takes place in  `swapBurnSend`, which calls `swapTokenForHat`, and currently looks like this:

```
    function swapBurnSend(uint256 _pid,
                        address _beneficiary,
                        uint256 _amountOutMinimum,
                        uint24[2] memory _fees) external onlyGovernance {
     ...
     uint256 hatsReceived = swapTokenForHAT(amount, token, _fees, _amountOutMinimum);
    ...
   }
 
    function swapTokenForHAT(uint256 _amount,
                            IERC20 _token,
                            uint24[2] memory _fees,
                            uint256 _amountOutMinimum)
    internal
    returns (uint256 hatsReceived)
    {
        if (address(_token) == address(HAT)) {
            return _amount;
        }
        require(_token.approve(address(uniSwapRouter), _amount), "HVE31");
        uint256 hatBalanceBefore = HAT.balanceOf(address(this));
        address weth = uniSwapRouter.WETH9();
        bytes memory path;
        if (address(_token) == weth) {
            path = abi.encodePacked(address(_token), _fees[0], address(HAT));
        } else {
            path = abi.encodePacked(address(_token), _fees[0], weth, _fees[1], address(HAT));
        }
        hatsReceived = uniSwapRouter.exactInput(ISwapRouter.ExactInputParams({
            path: path,
            recipient: address(this),
            deadline: block.timestamp,
            amountIn: _amount,
            amountOutMinimum: _amountOutMinimum
        }));
        require(HAT.balanceOf(address(this)) - hatBalanceBefore >= _amountOutMinimum, "HVE32");
    }
```

We propose to change this function by providing the option of routing the call to a contract and function call chosen by the caller of the function. I.e. the function signature of `swapBurnSend` and `swapTokenForHat` would change as follows:

```
    function swapBurnSend(uint256 _pid,
                        address _beneficiary,
                        uint256 _amountOutMinimum,
                       address _routingContract, // contract to call 
                        uint256 _routingPayLoad, // call to the _routingContract,
                        bytes _routingValue, // amount of eth to send to the _routingContract with the payload
                  
       ) {
    [...]
    uint256 hatsReceived = swapTokenForHAT(amount, token, _amountOutMinimum, _routingContract,_routingPayLoad, _routingValue );
    [...]
}

    function swapTokenForHAT(uint256 _amount,
                        IERC20 _token,
                        uint256 _amountOutMinimum, 
                        address _routingContract, // contract to call 
                        uint256 _routingPayLoad, // call to the _routingContract,
                        bytes _routingValue, // amount of eth to send to the _routingContract with the payload
     )
    internal
    returns (uint256 hatsReceived)
    {
        if (address(_token) == address(HAT)) {
            return _amount;
        }
        require(_token.approve(_routingContract, _amount), "HVE31");
        uint256 hatBalanceBefore = HAT.balanceOf(address(this));
        (bool success, bytes memory returndata) = _routingContract.call{value: _routingValue}(_routingPayLoad);
        require(success, "swap not successful");
        hatsReceived = HAT.balanceOf(address(this)) - hatBalanceBefore; // we check the actual amount received, instead of relying on uniswap to report the right value, as we did before
        require(hatsReceived >= _amountOutMinimum, "HVE32");
    }
```

What thsi will do:
* it's up to governance to decide where and how to sell the vault tokens
* Governance can call uniswapv3 by creating the payload of the call to `exactInput` off-chain. 
* We need to add anti-reentrancy modifiers, I think
* Governance cannot sell more tokens than allowed because it approves only the exact amount of tokens. 
* We may want to check as well if indeed all tokens that should be sold are being sold, i.e. to add a requirement that `tokenBalanceBefore - tokenBalanceAfter == amount`

ben-kaufman:
Fix #99


jellegerbrandy (dismissed):
This looks good to me. Can you resolve the conflicts that resulted from the refactor?

jellegerbrandy (approved):
Looks good to me :)

PR #146, opened 2022-05-06
reward -> bounty

lirona:


contracts/HATVaults.sol
```diff
@@ -1075,33 +1075,33 @@ contract  HATVaults is Governable, ReentrancyGuard {
...
     }
 
     // GET INFO for UI
     /**
-    * @dev getRewardPerBlock return the current pool reward per block
-    * @param _pid1 the pool id.
-    *        if _pid1 = 0 , it return the current block reward for whole pools.
-    *        otherwise it return the current block reward for _pid1-1.
+    * @dev Return the current pool reward per block
+    * @param _pid The pool id.
```
jellegerbrandy, line 1093: This `getRewardPerBlock` has a very shitty API, as it takes not "the poll Id" but "the pool id plus 1" as its main argument (which is why the argument is named so badly). I created #148 to handle that

jellegerbrandy (approved):

Issue #93, opened 2022-02-19
Migration from v1 to v2


jellegerbrandy:
As the contract is not upgradeable, we must find a way to migrate the funds and users. Ideally, this migration would require no action on the part of the user. 

## Migration process
The idea is for each pool we:

(A) pause the original pool 
(B) send the funds over to the new contract
(C) take a snapshot of the distribution of shares and create a pool on the new contact with the same settings and shares distribution as the old contract

1. Governance pauses deposits to the original pool by calling `setPool` with `_depositPause = false`.
2. (governance calls `setRewardsLevelsDelay(pid, 2 days)` to set the daly to  the minimum of 2 days , only if the current value is higher)
3. Governance calls `setPendingRewardsLevels(pid, [10000-1])` to create single rewards level that will pay out 99.99% of the pool
5.  Two days later, governance calls `setRewardsLevels()` 
5.   Governance calls `setRewardsSplit(pid, rewardsplit)` where `rewardsplit` sets all values to `0` except for `hackerReward` which should be set to `10000`
6. **Committee** creates a claim by calling `pendingApprovalClaim(pid, newVaultsAddress, 0)` (where 0 denotes the 99.99% severity level). 
7. At this point both deposits and withdrawals from the pool are pause. The migraiton script takes a snapshot of all shares (i.e. for each account it's `balance` and `rewardDebt`), to be used later
8. Governance calls `approveClaim(pid)` which will send 99.99% of the funds to the new contract 😅 
9. Governance creates a new pool on the new contract by calling `addPool(allocPoint, lpToken, rewardsLevels, rewardsSplit, descriptionHash, rewardVestingParams, rewardPerShare, balance, depositPause)` with the same  settings as the original pool
10. Governance transfers the shares distribution of the previous pool to the new one by calling a **new function** called `setShares(uint256 _pid, address[] accounts, uint256[] shares, uint256[] rewardDebts)`, reading the values from the original pool. If there are a lot of shares, this function perhaps needs to be called several times to stay within the gas limit  --> See specs of this new function below
11.  Governance unpauses the vault by calling `setVault(pid, ...)`
12. The **committee** calls `committeeCheckIn` on the new contract

### New arguments in `addPool`

We add new parameters to the `addPool` function which will now look like this:
```
    function addPool(uint256 _allocPoint,
                    address _lpToken,
                    address _committee,
                    uint256[] memory _rewardsLevels,
                    RewardsSplit memory _rewardsSplit,
[...]

PR #153, opened 2022-05-10
Prepare for migration

ben-kaufman:
Prepares for V2 migration based on the plan detailed in #93


contracts/HATVaults.sol
```diff
@@ -949,20 +939,48 @@ contract  HATVaults is Governable, ReentrancyGuard {
...
         poolDepositPause[_pid] = _depositPause;
         emit SetPool(_pid, _allocPoint, _registered, _descriptionHash);
     }
 
+    function setPoolInitialized(uint256 _pid) external onlyGovernance {
+        require(poolInfo.length > _pid, "HVE23");
+       poolInitialized[_pid] = true;
+    } 
+
+    function setShares(
```
lirona, line 953: should set poolInitialized to true here and not in a separate function, no?

contracts/HATVaults.sol
```diff
@@ -949,20 +939,48 @@ contract  HATVaults is Governable, ReentrancyGuard {
...
-        require(poolInfos.length > _pid, "HVE23");
+        require(poolInfo.length > _pid, "HVE23");
         set(_pid, _allocPoint);
         poolDepositPause[_pid] = _depositPause;
         emit SetPool(_pid, _allocPoint, _registered, _descriptionHash);
     }
 
+    function setPoolInitialized(uint256 _pid) external onlyGovernance {
+        require(poolInfo.length > _pid, "HVE23");
+       poolInitialized[_pid] = true;
```
jellegerbrandy, line 950: indentation

contracts/HATVaults.sol
```diff
@@ -949,20 +939,48 @@ contract  HATVaults is Governable, ReentrancyGuard {
...
+    external onlyGovernance {
+        require(!poolInitialized[_pid], "HVE38");
+        require(poolInfo.length > _pid, "HVE23");
+        require(_accounts.length == _shares.length, "HVE39");
+        require(_accounts.length == _rewardDebts.length, "HVE39");
+        PoolInfo storage pool = poolInfo[_pid];
+        pool.rewardPerShare = _rewardPerShare;
+        pool.balance = _balance;
+        for (uint256 i = 0; i < _accounts.length; i++) {
+            userInfo[_pid][_accounts[i]] = UserInfo({
```
jellegerbrandy, line 969: we must also update totalshares here, right?

scripts/migration-snapshot.js
```diff
@@ -0,0 +1,71 @@
+const { network } = require("hardhat");
+const fs = require("fs");
+const CONFIG = require("./addresses.js");
+const HATVaults = JSON.parse(
```
jellegerbrandy, line 4: this is the ABI right?

jellegerbrandy: mb name the file in a more clearer way (like hatvaultsv1ABI.json :)

jellegerbrandy (changes requested):

contracts/HATVaults.sol
ben-kaufman: Fixed

contracts/HATVaults.sol
ben-kaufman: Fixed

jellegerbrandy (approved):

PR #154, opened 2022-05-10
Max test coverage

### LINKED ISSUE #144: Have 100% test coverage

ben-kaufman:
Fix #144

Test coverage is impossible to get to 100% though because the HATToken has 3 require checks that are impossible to not meet - ie. check for impossible scenarios. 
We could update the HATToken code to remove them, or we could remove the HATToken from coverage count, but otherwise reaching 100% is impossible.

### DIFF
--- a/contracts/HATToken.sol
+++ b/contracts/HATToken.sol
@@ -245,7 +245,7 @@ contract HATToken is IERC20 {
         bytes32 structHash = keccak256(abi.encode(PERMIT_TYPEHASH, owner, spender, rawAmount, nonces[owner]++, deadline));
         bytes32 digest = keccak256(abi.encodePacked("\x19\x01", domainSeparator, structHash));
         address signatory = ecrecover(digest, v, r, s);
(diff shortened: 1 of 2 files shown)


jellegerbrandy:
We need to use `assert` instead of require for impossible-to-meet conditions, and I guess then we wil get to 100%

contracts/HATToken.sol
```diff
@@ -485,7 +485,7 @@ contract HATToken is IERC20 {
 
     function add96(uint96 a, uint96 b, string memory errorMessage) internal pure returns (uint96) {
         uint96 c = a + b;
-        require(c >= a, errorMessage);
+        assert(c >= a);
```
jellegerbrandy, line 488: We have a linter warning `errorMessage` is unused - can you remove the argument?

jellegerbrandy (changes requested):
Very nice just one small remarkt

jellegerbrandy (approved):

PR #157, opened 2022-05-12
Create HAT Diamond [keep this for reference]

ben-kaufman:
Use the Diamond pattern for the HATVaults contract


scripts/hatvaultsdeploy.js
```diff
@@ -1,80 +1,169 @@
...
     console.warn(
       "You are trying to deploy a contract to the Hardhat Network, which" +
         "gets automatically created and destroyed every time. Use the Hardhat" +
         " option '--network localhost'"
     );
   }
 
-  const addresses = ADDRESSES[network.name];
+  const diamond = require("./diamond-util.js");
+  let totalGasUsed = ethers.BigNumber.from("0");
```
jellegerbrandy, line 22: ?

contracts/facets/DepositFacet.sol
```diff
@@ -0,0 +1,62 @@
+// SPDX-License-Identifier: MIT
```
jellegerbrandy, line 1: seems a bit weird to me to have a deposit and a separate reward facet, let's merge thos

contracts/facets/CommitteeFacet.sol
```diff
@@ -0,0 +1,82 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.6;
```
jellegerbrandy, line 2: Seems to me the "setFoo" functions should in the paramsfacet, right?

jellegerbrandy (changes requested):
I think this is very good and clear, I left some small suggestions

Heads up that will ahve a tough time fixing merge conflicts later, so I did not check if all the code was propertly transfered yet, I'll do another review then

contracts/facets/CommitteeFacet.sol
ben-kaufman: Well these which are set by the committee I put here, and the general params set by hats governance I put in the params. Makes sense or what do you think?

contracts/facets/DepositFacet.sol
ben-kaufman: Ok merged them

scripts/hatvaultsdeploy.js
ben-kaufman: Removed

Issue #127, opened 2022-04-13
safeTransferReward now fails silently if there are not enough tokens to pay out the reward - need to find a better solution


lirona:
at the moment if there are not enough HAT tokens to transfer the whole requested reward, the function transfers the available amount and finishes. Should fail and not transfer any tokens in this case (or update user's rewardDebt so that she gets the rest next time it's possible)

lirona:
change the event

jellegerbrandy:
So here is an alternative type of solution.
The problem we have is that rewards cannot be paid out because the HATs balance is insufficient to pay out the amount... instead of waiting for that to happen, we can "fail" earlier -namely when we call `updatePool`, when we calculate the rewardPErShare we also take into account the available of HATs to paya that reward. I.e. if we set the `rewardPerShare` value in such a way tha `rewardPerShare * totalShares <= hatsBalance`

jellegerbrandy:
This last solution will not work, what we need is an equation of the type `rewardPerShare * totalShares <= hatsBalance + hatsPaidOut`, which would require a `hatsPaidOut` counter. It also has the drawback that not all pools are treated equally.

jellegerbrandy:
It seems to me thsi was was also a problem in the v1 version of HatVaults (which fails silently if there are not enough tokens to pay out the reward)
The original pancakeswap implementation https://github.com/pancakeswap/pancake-farm/blob/master/contracts/MasterChef.sol#L222 does not have a limit on minting, I think, so does not have the probelm)
By way of solutions, I see the following:
1- we do what we do now, namely "fail silently" and simply stop paying the rewards to users on withdrawal and deposit (they have no way to recover that)
2- we add more bookkeeping, for example a counter saying how much rewards we already ahve paid out, so we can manage the reward program
3- we can actually throw an error when rewards cannot be paid out on deposit and reward (so users can only emergenyWithdraw)

At this point, I would be for (3). The reasoning here is that this is a real error: HAts is responsible for funding the contract with enough HATs to pay out the rewards, if it does not, then governance either needs to provide enough HATs, or otherwise change the reward program (i.e. call setRewardMultiplier) so that it is not promised too much

PR #163, opened 2022-05-18
saner test setup

lirona:


contracts/HATVaults.sol
```diff
@@ -1118,14 +1129,14 @@ contract  HATVaults is Governable, ReentrancyGuard {
...
-    function pendingReward(uint256 _pid, address _user) external view returns (uint256) {
+    function pendingReward(uint256 _pid, address _user) public view returns (uint256) {
         PoolInfo storage pool = poolInfos[_pid];
         UserInfo storage user = userInfo[_pid][_user];
         uint256 rewardPerShare = pool.rewardPerShare;
 
         if (block.number > pool.lastRewardBlock && pool.totalShares > 0) {
             uint256 reward = calcPoolReward(_pid, pool.lastRewardBlock, globalPoolUpdates.length-1);
-            rewardPerShare += (reward * 1e12 / pool.totalShares);
+            rewardPerShare += (reward * 1e12 / pool.totalShares);//TODO we don't update the pool's rewardPerShare here!!
```
lirona, line 1139: @jellegerbrandy why not update it while we're at it?

contracts/HATVaults.sol
```diff
@@ -1118,7 +1129,7 @@ contract  HATVaults is Governable, ReentrancyGuard {
         }
     }
 
-    function pendingReward(uint256 _pid, address _user) external view returns (uint256) {
+    function pendingReward(uint256 _pid, address _user) public view returns (uint256) {
         PoolInfo storage pool = poolInfos[_pid];
         UserInfo storage user = userInfo[_pid][_user];
         uint256 rewardPerShare = pool.rewardPerShare;
```
lirona, line 1135: @jellegerbrandy why not update the pool's reward per share if we enter the next if block?
we calculate already.. can update lastRewardBlock so it will not update next

lirona: actually what is this used for? if it's only for the user to check for himself then this should use msg.sender, and ok to not update

contracts/HATVaults.sol
```diff
@@ -496,19 +492,34 @@ contract  HATVaults is Governable, ReentrancyGuard {
...
+    /**
+    * @dev Transfer the the pending HAT reward to the user, as much as is available in the contract
+    * @param _pid The pool id
+    * @param _user The address of the user
+    * @return amountNotRewarded - amount of HAT reward left to reward user (0 if all pending reward was transferred)
+    */
+    function safeTransferReward(uint256 _pid, address _user) internal returns (uint256 amountNotRewarded) {
+        PoolInfo storage pool = poolInfos[_pid];
+        UserInfo storage user = userInfo[_pid][_user];
+        uint256 pendingReward = uint256(int256(user.shares * pool.rewardPerShare / 1e12) - user.rewardDebt);
```
jellegerbrandy, line 508: so this will fail if shares is 0 and rewarddebt is positive..

jellegerbrandy (changes requested):

contracts/HATVaults.sol
jellegerbrandy: so I guess this is a convenience function for the ui

jellegerbrandy:
Ok, so I think your approach with trying to use `rewardDebt` will not work. We could do a similar approach but we need a track  `user.rewardWeCouldNotPayOut`. separately.

I  updated the issue https://github.com/hats-finance/hats-contracts/issues/127  with an alternative idea

In any case, regarding this PR, I think you should START writing tests :)

contracts/HATVaults.sol
```diff
@@ -85,7 +85,7 @@ contract  HATVaults is Governable, ReentrancyGuard {
 
     struct UserInfo {
         uint256 shares;     // The user share of the pool based on the shares of lpToken the user has provided.
-        uint256 rewardDebt; // Reward debt. See explanation below.
+        int256 rewardDebt; // Reward debt. See explanation below.
```
jellegerbrandy, line 88: booh

PR #164, opened 2022-05-18
Fix #159

### LINKED ISSUE #158: emergencyWithdraw should be non-reenetrant or implement checks-effects-interactions pattern
theoretically it is vulnerable to reentrancy, so lets just follow best practices here

### LINKED ISSUE #159: in _deposit, do not assume that the call to transfer does transfer exactly the right amount of tokens
instead, check the difference in balance

ben-kaufman:
Fix #159
Also fixes  #158


contracts/HATVaults.sol
```diff
@@ -485,7 +486,9 @@ contract  HATVaults is Governable, ReentrancyGuard {
         }
         if (_amount > 0) { // will only be 0 in case of claimReward
             uint256 lpSupply = pool.balance;
+            uint256 balanceBefore = pool.lpToken.balanceOf(address(this));
             pool.lpToken.safeTransferFrom(address(msg.sender), address(this), _amount);
+            require(pool.lpToken.balanceOf(address(this)) - balanceBefore == _amount, "HVE45");
```
lirona, line 491: we don't want to disable deposits in case there is a fee taken or something like that, just to check how much was actually transferred and update everything accordingly

lirona: same for withdrawal

jellegerbrandy: exactly, replace this line with `_amount = pool.lpToken.balanceOf(address(this)) - balanceBefore` and I guess we are done

jellegerbrandy: NOT the same for withdrawal, though, with withdrawal I guess the logic can be left unchanged

jellegerbrandy (changes requested):
See the comments:
- do not fail when deposit is taxed, just measure how much is actually deposit
- when withdrawing, we do not need any addditional checks afaics

contracts/vaults/Withdraw.sol
```diff
@@ -60,12 +57,18 @@ contract Withdraw is Swap {
...
         emit EmergencyWithdraw(msg.sender, _pid, factoredBalance);
     }
+
+    function safeWithdrawPoolToken(IERC20 _lpToken, uint256 _totalAmount, uint256 _fee) internal {
+        uint256 balanceBefore = _lpToken.balanceOf(address(this));
+        if (_fee > 0) {
+            _lpToken.safeTransfer(governance(), _fee);
+        }
+        _lpToken.safeTransfer(msg.sender, _totalAmount - _fee);
+        require(balanceBefore - _lpToken.balanceOf(address(this)) == _totalAmount, "HVE45");
```
jellegerbrandy, line 70: So I do not think we need this require. I mean, I can think of 2 scenarios here: 
- either the require fails for a good reason, in which case our users cannot withdraw, and it is our fault
- the require fails for a bad reason, in which case the token is problematic, and it can fail in many other ways, perhaps much worse, and this require is not saving us from any problem
So my opinion we should drop this requirement

jellegerbrandy (changes requested):

PR #165, opened 2022-05-19
move some functions to library to see what if we can diminish code size that way [experiment, do not merge]

jellegerbrandy:
This uis a PoC, it brings down the side of the HATVaults contract from 23.969 (current development branch) to 23.871

### DIFF
--- a/contracts/HATVaults.sol
+++ b/contracts/HATVaults.sol
@@ -12,7 +12,7 @@ import "./Governable.sol";
 import "./HATToken.sol";
 import "./tokenlock/ITokenLockFactory.sol";
 import "./interfaces/ISwapRouter.sol";
-
+import {HATVaultsLib} from "./HATVaultsLib.sol";
 
 // Errors:
 // HVE01: Only committee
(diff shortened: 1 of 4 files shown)


contracts/HATVaults.sol
```diff
@@ -65,24 +65,6 @@ import "./interfaces/ISwapRouter.sol";
...
-        uint256 setBountyLevelsDelay;
-        // period of time after withdrawRequestPendingPeriod where it is possible to withdraw
-        // (after which withdrawal is not possible)
-        uint256 withdrawRequestEnablePeriod;
-        // period of time that has to pass after withdraw request until withdraw is possible
-        uint256 withdrawRequestPendingPeriod;
-        uint256 claimFee;  //claim fee in ETH
-    }
-
     struct UserInfo {
```
lirona, line 68: why is this struct in both files?

contracts/HATVaults.sol
```diff
@@ -1146,44 +1079,15 @@ contract  HATVaults is Governable, ReentrancyGuard {
...
-        totalBountyAmount * bountyInfos[_pid].bountySplit.swapAndBurn
-        / (HUNDRED_PERCENT * HUNDRED_PERCENT);
-        claimBounty.governanceHat =
-        totalBountyAmount * bountyInfos[_pid].bountySplit.governanceHat
-        / (HUNDRED_PERCENT * HUNDRED_PERCENT);
-        claimBounty.hackerHat =
-        totalBountyAmount * bountyInfos[_pid].bountySplit.hackerHat
-        / (HUNDRED_PERCENT * HUNDRED_PERCENT);
+    returns(HATVaultsLib.ClaimBounty memory claimBounty) {
+      return HATVaultsLib.calcClaimBounty(_pid, _severity, poolInfos, bountyInfos[_pid]);
```
lirona, line 1083: why pass all the poolInfos and not just poolInfos[_pid]?

lirona: or just the balance since it's the only thing needed? and then it can be pure

PR #166, opened 2022-05-19
Split HATVaults into files

ben-kaufman:
Commnet: Somehow this seems to make the HATVaults lighter but I'm not sure why


jellegerbrandy (approved):
great. I unlinearized the import tree so it looks more modular (I think it compiles the same way)

PR #168, opened 2022-05-20
rename facet files

jellegerbrandy:


lirona:
the bounty params are split between 2 files: committee and params, i think they should be one file. or change the name to governance and committee because that's what they are split according to

PR #174, opened 2022-05-24
lint js files and some cleanup

jellegerbrandy:


lirona:
so the solidity style guide recommends maximum line length of 79 or 99. i think 99 is better because it's not more readable
https://docs.soliditylang.org/en/latest/style-guide.html#maximum-line-length

lirona (commented):
so just think it's too narrow, but that's just me

jellegerbrandy:
ok so I do not want to get into the "these are the best linting rules" discussion, so I'll ignore your comments :P

PR #177, opened 2022-05-25
reward transfers to fail if not enough balance

lirona:


contracts/vaults/Swap.sol
```diff
@@ -62,23 +62,24 @@ contract Swap is Base {
...
     returns (uint256 swapTokenReceived)
     {
         if (address(_token) == address(swapToken)) {
             return _amount;
         }
-
+        require(whitelistedRouters[_routingContract], "HVE44");
         require(_token.approve(_routingContract, _amount), "HVE31");
         uint256 balanceBefore = swapToken.balanceOf(address(this));
+        require(balanceBefore >= _amountOutMinimum, "HVE45");
```
jellegerbrandy, line 78: ?????

contracts/vaults/Base.sol
```diff
@@ -290,15 +290,17 @@ contract  Base is Governable, ReentrancyGuard {
...
+    * @param _to The address to transfer the reward to
+    * @param _amount The amount of rewards to transfer
+    * @param _pid The pool id
+   */
     function safeTransferReward(address _to, uint256 _amount, uint256 _pid) internal {
-        if (_amount > rewardAvailable) { 
-            _amount = rewardAvailable; 
-        }
-        rewardAvailable -= _amount;
+        require(rewardToken.balanceOf(address(this)) >= _amount, "HVE46");
```
jellegerbrandy, line 301: Please do not do this The balance of HATs is (a) the amount reservered for rewards and (b) the amount deposited in the vaults that hold HATs. With this change, you are paying rewards from a vault.

jellegerbrandy (changes requested):

contracts/vaults/Base.sol
```diff
@@ -383,22 +385,22 @@ contract  Base is Governable, ReentrancyGuard {
 
     function validateSplit(BountySplit memory _bountySplit) internal pure {
         require(_bountySplit.hackerVested
-            + _bountySplit.hacker
-            + _bountySplit.committee
-            + _bountySplit.swapAndBurn
-            + _bountySplit.governanceHat
-            + _bountySplit.hackerHat == HUNDRED_PERCENT,
-        "HVE29");
+        + _bountySplit.hacker
```
jellegerbrandy, line 388: this is not an improvement

contracts/vaults/Base.sol
```diff
@@ -383,22 +385,22 @@ contract  Base is Governable, ReentrancyGuard {
...
 
     function getDefaultBountySplit() public pure returns (BountySplit memory) {
         return BountySplit({
-            hackerVested: 6000,
-            hacker: 2000,
-            committee: 500,
-            swapAndBurn: 0,
-            governanceHat: 1000,
-            hackerHat: 500
+            hackerVested : 6000,
```
jellegerbrandy, line 398: this is not the spacing we use in other parts of the code

jellegerbrandy (changes requested):

contracts/vaults/Base.sol
lirona: but they are deposited into the vault also..
https://github.com/hats-finance/hats-contracts/blob/e3b6ae9d16a8babfa4ce647e15acf41f3e2efcf9/contracts/vaults/Deposit.sol#L23

contracts/vaults/Swap.sol
lirona: checking that we have enough rewards to do the swap, what's wrong with that?

contracts/vaults/Base.sol
```diff
@@ -290,15 +290,17 @@ contract  Base is Governable, ReentrancyGuard {
         pool.lastProcessedTotalAllocPoint = lastPoolUpdate;
     }
 
-    // Safe rewardToken transfer function, transfer HATs from the contract only if they are earmarked for rewards
+    /**
+    * @dev Safe HAT transfer function, transfer rewards from the contract only if there are enough
```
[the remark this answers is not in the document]
lirona: still i think internal funcs should be documented, maybe no need for natspec though, you're right

contracts/vaults/Base.sol
lirona: i think it's good that it's clear, but not necessary, can remove

contracts/vaults/Base.sol
lirona: i got a lint warning for this before changing

jellegerbrandy: well you probably have a problem with your linter then

contracts/vaults/Base.sol
jellegerbrandy: the linter should decide this anyway

contracts/vaults/Base.sol
lirona: :)
ok i'll put it back

contracts/vaults/Base.sol
lirona: on variable names?..

lirona: oh you mean the identation, not changing `hackerHat` to `hackerHatVested`?

jellegerbrandy: yess

contracts/vaults/Deposit.sol
```diff
@@ -7,8 +7,11 @@ contract Deposit is Base {
     using SafeERC20 for IERC20;
 
     function depositReward(uint256 _amount) external {
+        uint256 balanceBefore = swapToken.balanceOf(address(this));
         rewardToken.transferFrom(address(msg.sender), address(this), _amount);
-        emit DepositReward(_amount, address(rewardToken));
+        uint256 rewardTokenReceived = swapToken.balanceOf(address(this)) - balanceBefore;
```
jellegerbrandy, line 12: I like the balance checking, but you should use the same token

jellegerbrandy (changes requested):
plus merge conflicts

contracts/vaults/Deposit.sol
lirona: oops! thanks

contracts/vaults/Deposit.sol
```diff
@@ -7,9 +7,11 @@ contract Deposit is Base {
     using SafeERC20Upgradeable for IERC20Upgradeable;
 
     function depositReward(uint256 _amount) external {
-        rewardAvailable += _amount;
+        uint256 balanceBefore = rewardToken.balanceOf(address(this));
         rewardToken.transferFrom(address(msg.sender), address(this), _amount);
-        emit DepositReward(_amount, address(rewardToken));
+        uint256 rewardTokenReceived = rewardToken.balanceOf(address(this)) - balanceBefore;
+        rewardAvailable += rewardTokenReceived;
+    emit DepositReward(_amount, rewardTokenReceived, address(rewardToken));
```
jellegerbrandy, line 14: you have to fix your linter!

contracts/vaults/Base.sol
```diff
@@ -289,15 +297,18 @@ contract  Base is OwnableUpgradeable, ReentrancyGuardUpgradeable {
...
     function safeTransferReward(address _to, uint256 _amount, uint256 _pid) internal {
-        if (_amount > rewardAvailable) { 
-            _amount = rewardAvailable; 
-        }
-        rewardAvailable -= _amount;
+        require(rewardAvailable >= _amount, "HVE46");
         rewardToken.transfer(_to, _amount);
-        // TODO: fix return of the requested amount
-        emit SafeTransferReward(_to, _pid, _amount, _amount);
+        rewardAvailable -= _amount;
```
jellegerbrandy, line 310: but why did you change the order here?
So in general there is a "best practice" rule to do state changes before external calls, to avoid reentrancy attacks. (In this case, the transfer poses no risk, and anyway we call safetransfer in other contexts, but it is still a good pattern to follow)

jellegerbrandy (changes requested):

contracts/vaults/Deposit.sol
lirona: i fixed this! refresh

lirona: oops i see i didn't commit

jellegerbrandy (approved):

PR #178, opened 2022-05-26
Upgradeable HATVaults

ben-kaufman:
Prepare hat vaults for upgradeability


.openzeppelin/unknown-31337.json
```diff
@@ -0,0 +1,2222 @@
+{
```
jellegerbrandy, line 1: committed by mistake?

contracts/vaults/Claim.sol
```diff
@@ -4,7 +4,7 @@ pragma solidity 0.8.6;
 import "./Base.sol";
 
 contract Claim is Base {
-    using SafeERC20 for IERC20;
+    using SafeERC20Upgradeable for IERC20Upgradeable;
```
jellegerbrandy, line 7: why? we are calling that token, we do not care if it upgradeable or not, or do we?

contracts/vaults/Pool.sol
```diff
@@ -65,7 +65,7 @@ contract Pool is Base {
             }));
         }
         poolInfos.push(PoolInfo({
-            lpToken: IERC20(_lpToken),
+            lpToken: IERC20Upgradeable(_lpToken),
```
jellegerbrandy, line 68: cam just be IERC20

scripts/hatvaultsdeploy.js
```diff
@@ -1,77 +1,84 @@
...
-  console.log("Deploying the contracts with the account:", deployerAddress);
-  console.log("Account balance:", (await deployer.getBalance()).toString());
+  if (!silent) {
+    console.log("Deploying the contracts with the account:", deployerAddress);
+    console.log("Account balance:", (await deployer.getBalance()).toString());
+  }
 
   const HATVaults = await ethers.getContractFactory("HATVaults");
-  const hatVaults = await HATVaults.deploy(
+  const hatVaults = await upgrades.deployProxy(HATVaults, [
```
jellegerbrandy, line 39: we should also decide on which proxy pattern to use: transparent or uups

It's a setting: https://docs.openzeppelin.com/upgrades-plugins/1.x/api-hardhat-upgrades

jellegerbrandy (changes requested):
some small changes.

@ben-kaufman do you have an opinion on transparentproxy vs the uups one?

ben-kaufman:
> @ben-kaufman do you have an opinion on transparentproxy vs the uups one?
UUPS is nice but it will make our HATVaults even heavier. Right now it's already about 1MB too big to deploy. So unless we can significantly reduce contract size we don't really have the option of UUPS.

contracts/vaults/Pool.sol
ben-kaufman: It gives an error deploying:
Error: Contract `HATVaults` is not upgrade safe

@openzeppelin/contracts/utils/Address.sol:191: Use of delegatecall is not allowed

Also even if can ignore the error I think it's best practice to use only the upgradeable repo with upgradeable contracts

jellegerbrandy: The error must be a bug, right?  In any case, I appreciete your point about using just a single dependency

PR #182, opened 2022-05-26
add _descriptionHash to SubmitClaim

lirona:


jellegerbrandy (changes requested):
this is all good. We should also have a test for this event though. The only place in the tests where this event is tested is on line 1751 - we should test there also if actually the descirption hash is part of the event (just add a line below)

lirona:
> this is all good. We should also have a test for this event though. The only place in the tests where this event is tested is on line 1751 - we should test there also if actually the descirption hash is part of the event (just add a line below)

link to line please? :)

lirona:
> this is all good. We should also have a test for this event though. The only place in the tests where this event is tested is on line 1751 - we should test there also if actually the descirption hash is part of the event (just add a line below)

not the right line..

jellegerbrandy (approved):

PR #196, opened 2022-05-30
Fix #184

### LINKED ISSUE #184: merge `deposit` and `_deposit` into a single function

ben-kaufman:
Fix #184

### DIFF
--- a/contracts/vaults/Deposit.sol
+++ b/contracts/vaults/Deposit.sol
@@ -35,7 +35,7 @@ contract Deposit is Base {
      * @param _pid The pool id
      */
     function claimReward(uint256 _pid) external {
-        _deposit(_pid, 0);
+        _claimReward(_pid);
         emit ClaimReward(_pid);
     }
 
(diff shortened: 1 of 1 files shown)


contracts/vaults/Deposit.sol
```diff
@@ -49,37 +50,35 @@ contract Deposit is Base {
...
+            userShares = pool.totalShares * transferredAmount / lpSupply;
+        }
+        user.shares += userShares;
+        pool.totalShares += userShares;
+        user.rewardDebt = user.shares * pool.rewardPerShare / 1e12;
         emit Deposit(msg.sender, _pid, _amount, transferredAmount);
     }
 
-    function _deposit(uint256 _pid, uint256 _amount) internal nonReentrant returns(uint256 _transferredAmount) {
+    function _claimReward(uint256 _pid, UserInfo memory _user) internal nonReentrant {
```
jellegerbrandy, line 73: what about moving the nonReentrant decorator to `deposit` and `withdraw` instead of hiding it here?

jellegerbrandy (changes requested):

contracts/vaults/Deposit.sol
```diff
@@ -49,37 +50,35 @@ contract Deposit is Base {
         require(_amount >= MINIMUM_DEPOSIT, "HVE27");
         //clear withdraw request
         withdrawEnableStartTime[_pid][msg.sender] = 0;
-        uint256 transferredAmount = _deposit(_pid, _amount);
+        PoolInfo storage pool = poolInfos[_pid];
+        UserInfo storage user = userInfo[_pid][msg.sender];
+        _claimReward(_pid, user);
+        uint256 lpSupply = pool.balance;
+        uint256 balanceBefore = pool.lpToken.balanceOf(address(this));
+        pool.lpToken.safeTransferFrom(address(msg.sender), address(this), _amount);
```
lirona, line 58: msg.sender is already an address :) we can remove the "address( )" around it

lirona: right? is there a meaning to this? i checked and runs the same

jellegerbrandy (approved):

PR #207, opened 2022-06-01
200 use a claimid instead of poolid

lirona:


contracts/vaults/Claim.sol
```diff
@@ -33,43 +33,52 @@ contract Claim is Base {
...
+    {
         require(_beneficiary != address(0), "HVE04");
         // require we are in safetyPeriod
         // solhint-disable-next-line not-rely-on-time
         require(block.timestamp % (generalParameters.withdrawPeriod + generalParameters.safetyPeriod) >=
         generalParameters.withdrawPeriod, "HVE05");
         require(_bountyPercentage <= bountyInfos[_pid].maxBounty, "HVE06");
-
-        submittedClaims[_pid] = SubmittedClaim({
+        uint256 claimId = uint256(keccak256(abi.encodePacked(_pid, block.number)));
```
lirona, line 44: wondering if should check that its not 0? (and then do uint256(keccak256(abi.encodePacked(_pid+1, block.number)) or something like that)
and also to check if exists already?

contracts/vaults/Claim.sol
```diff
@@ -93,29 +102,34 @@ contract Claim is Base {
...
-                        submittedClaim.beneficiary,
-                        submittedClaim.bountyPercentage,
-                        tokenLock,
-                        claimBounty);
-        assert(poolInfos[_pid].balance > 0);
+        swapAndBurns[pid] += claimBounty.swapAndBurn;
+        governanceHatRewards[pid] += claimBounty.governanceHat;
+        hackersHatRewards[submittedClaim.beneficiary][pid] += claimBounty.hackerHatVested;
+        delete activeClaims[pid];
+        delete submittedClaims[_claimId];
```
lirona, line 109: is there a reason to keep this maybe? and check activeClaims in approve and dismiss?

contracts/vaults/Base.sol
```diff
@@ -274,7 +279,7 @@ contract Base is OwnableUpgradeable, ReentrancyGuardUpgradeable {
     }
 
     modifier noSubmittedClaims(uint256 _pid) {
```
jellegerbrandy, line 281: maybe also rename the modifier at this point to "noActiveClaims"

contracts/vaults/Claim.sol
```diff
@@ -33,43 +33,58 @@ contract Claim is Base {
...
         // require we are in safetyPeriod
         // solhint-disable-next-line not-rely-on-time
         require(block.timestamp % (generalParameters.withdrawPeriod + generalParameters.safetyPeriod) >=
         generalParameters.withdrawPeriod, "HVE05");
         require(_bountyPercentage <= bountyInfos[_pid].maxBounty, "HVE06");
-
-        submittedClaims[_pid] = SubmittedClaim({
+        uint256 claimId;
+        uint256 nonce;
+        do
```
jellegerbrandy, line 46: I see what you are doing but I think it is too complex.
We will only have a collision with ids if **within a single block** the following happens:

- a claim is submitted and gets id keccak(poolId, block.number)
- that claim is resolved
- a new claim is submitted <- this would get the same id as the first claim

IT is almost impossible to have this situation, and if it occurs probably something went wrong. So I do not think we need to write code especially to resolve this situation.

Instead, I'd suggest:
```
claimId = keccack256(_pid, block.number)
require(submittedClaims[claimId].beneficiary != 0, "a claim was already submitted in thsi block")
```

contracts/vaults/Claim.sol
```diff
@@ -33,43 +33,58 @@ contract Claim is Base {
...
-                            + claimBounty.swapAndBurn
-                            + claimBounty.hackerHatVested
-                            + claimBounty.governanceHat;
+    function approveClaim(uint256 _claimId) external onlyOwner nonReentrant {
+        SubmittedClaim memory submittedClaim = submittedClaims[_claimId];
+        require(submittedClaim.beneficiary != address(0), "HVE10");
+        uint256 pid = submittedClaim.pid;
+        BountyInfo storage bountyInfo = bountyInfos[pid];
+        IERC20Upgradeable lpToken = poolInfos[pid].lpToken;
+        ClaimBounty memory claimBounty = calcClaimBounty(pid, submittedClaim.bountyPercentage);
```
jellegerbrandy, line 81: not really part of this pr but here "memory" can be replaced by "storage" to save some gas

jellegerbrandy: i think :)

contracts/vaults/Claim.sol
```diff
@@ -33,43 +33,58 @@ contract Claim is Base {
...
-        ClaimBounty memory claimBounty = calcClaimBounty(_pid, submittedClaim.bountyPercentage);
-        poolInfos[_pid].balance -= claimBounty.hacker
-                            + claimBounty.hackerVested
-                            + claimBounty.committee
-                            + claimBounty.swapAndBurn
-                            + claimBounty.hackerHatVested
-                            + claimBounty.governanceHat;
+    function approveClaim(uint256 _claimId) external onlyOwner nonReentrant {
+        SubmittedClaim memory submittedClaim = submittedClaims[_claimId];
+        require(submittedClaim.beneficiary != address(0), "HVE10");
```
jellegerbrandy, line 77: mb it is more straightforward to require here that `activeClaims[_claimId]`

(should come down to the same thing I guess, but seems more readable)

contracts/vaults/Claim.sol
jellegerbrandy: Well, it saves some gas. And the data will remain available to offchain apps bc we have the events, so there is no real need to keep it

contracts/vaults/Claim.sol
```diff
@@ -93,29 +108,34 @@ contract Claim is Base {
...
     */
-    function dismissClaim(uint256 _pid) external {
+    function dismissClaim(uint256 _claimId) external {
+        uint256 pid = submittedClaims[_claimId].pid;
         // solhint-disable-next-line not-rely-on-time
-        require(msg.sender == owner() || submittedClaims[_pid].createdAt + 5 weeks < block.timestamp, "HVE09");
-        delete submittedClaims[_pid];
-        emit DismissClaim(_pid);
+        require(msg.sender == owner() || submittedClaims[_claimId].createdAt + 5 weeks < block.timestamp, "HVE09");
+        require(submittedClaims[_claimId].beneficiary != address(0), "HVE10");
```
jellegerbrandy, line 135: So this check is not really necessary, bc if beneficiary = 0, then the claim was already dismissed or never existed in the firsts place.
That said, I like to have an explicit check here, so let's just keep it

jellegerbrandy (changes requested):
Nice! I have just some small changes to suggest

contracts/vaults/Base.sol
lirona: i was thinking to change the struct and mapping too

contracts/vaults/Claim.sol
lirona: that will actually be totally impossible with the challenge period.. i was thinking more about hash collisions, almost impossible as they may be. that's why i was checking for 0 too

contracts/vaults/Claim.sol
lirona: it is :) will change

contracts/vaults/Claim.sol
lirona: i was worried it could be unsafe because that way approveClaim(0) would pass

contracts/vaults/Claim.sol
lirona: i think we have to check because if we dont then if this is submitted with an invalid claimId it would dismiss the claim in pool 0..

contracts/vaults/Claim.sol
jellegerbrandy: ah sorry it is me who was confused :)

contracts/vaults/Claim.sol
jellegerbrandy: ah no we do not worry about hash collisions
in theory the challenge period can be set to 0, and so, again theoretcially, I think my scenario could happen, right?

lirona: hmm even when it's set to 0 the submitting of a claim and approving/dismissing can't happen in the same block because they have to see the claim before approving/dismissing and it can't happen in the same time

lirona: but hash collisions could theoretically happen :)

jellegerbrandy (approved):

PR #212, opened 2022-06-05
documentation in HATVaults.sol and remove hardhat test from pipeline.yml

lirona:


.github/workflows/pipeline.yml
```diff
@@ -28,9 +28,6 @@ jobs:
       - name: Compile contracts
```
lirona, line 28: we can remove this too because hardhat test compiles before each run

jellegerbrandy (approved):

PR #213, opened 2022-06-05
remove default bounty split and add check for hacker payout

lirona:


contracts/vaults/Base.sol
```diff
@@ -337,6 +326,7 @@ contract Base is OwnableUpgradeable, ReentrancyGuardUpgradeable {
     }
 
     function validateSplit(BountySplit memory _bountySplit) internal pure {
+        require(_bountySplit.hacker + _bountySplit.hackerVested != 0, "HVE47");
```
jellegerbrandy, line 329: you added this? It's not necessarily wrong but it is not very useful at all. We could set _bountySplit.hacker to 1 wei  (so for all pracctical purposes it is 0) and this check would pass. I say remove :)

jellegerbrandy (changes requested):

jellegerbrandy (approved):

PR #238, opened 2022-06-17
Fix #225

### LINKED ISSUE #225: rewardDebt is not updated in claimReward
The _claimREward function is missing the following line: 

`user.rewardDebt = user.shares * pool.rewardPerShare / 1e12;`

These are two bugs, really:

- claimReward is broken
- there is no test for this behavior (in this case, a test calling claimReward two times in a row and checking if the values are as expected)

### LINKED ISSUE #231: reentrancy worries (audit #5)
> Adam Kolar | g0, [09/06/2022 16:48]
> [In reply to Hatter]
> 5) So malicious token contract could be actually used to claim reward multiple times, because rewardDebt is updated only after the token contract is called in deposit and withdraw and the reward is claimed before. It's important to update userInfo[_pid][msg.sender].rewardDebt as soon as reward is transferred or ideally even before it is transferred in Deposit._claimReward this will also solve issue 1)
> 
> Adam Kolar | g0, [09/06/2022 16:52]
> actually the issue is mitigated because deposit and withdraw` functions are nonReentrant but it's still exploitable because claimReward is not nonReentrant
> 
> Adam Kolar | g0, [09/06/2022 16:55]
> Also ideally, user.rewardDebt in Deposit.deposit and Withdraw.withdraw should be incremented and decremented not set, otherwise reentrancy can be used to gain shares without rewardDebt being updated because the updates will be overwritten by the outermost call. This issue is not exploitable due to the nonReentrant modifier, but I would still prefer the implementation to be adjusted.
> 
> Adam Kolar | g0, [09/06/2022 16:58]
> Also to prevent all possible reentrancy issues emergencyWithdraw should be also nonReentrant

ben-kaufman:
Fix #225

The test was broken because the hat vaults was configured in a way that the reward was 0

This should also fix #231 now.

### DIFF
--- a/contracts/vaults/Deposit.sol
+++ b/contracts/vaults/Deposit.sol
@@ -40,7 +40,6 @@ contract Deposit is Base {
 
         user.shares += userShares;
         pool.totalShares += userShares;
-        user.rewardDebt = user.shares * pool.rewardPerShare / 1e12;
 
         emit Deposit(msg.sender, _pid, _amount, transferredAmount);
     }
@@ -97,6 +96,7 @@ contract Deposit is Base {
         // if the user already has funds in the pool, give the previous reward
         if (_user.shares > 0) {
             uint256 pending = _user.shares * poolInfos[_pid].rewardPerShare / 1e12 - _user.rewardDebt;
+            userInfo[_pid][msg.sender].rewardDebt += pending;
             if (pending > 0) {
                 safeTransferReward(msg.sender, pending, _pid);
             }
--- a/contracts/vaults/Deposit.sol
+++ b/contracts/vaults/Deposit.sol
@@ -25,7 +25,6 @@ contract Deposit is Base {
         updatePool(_pid);
         if (user.shares > 0) {
             uint256 pending = user.shares * pool.rewardPerShare / 1e12 - user.rewardDebt;
-            // no need to update the rewardDebt yet because this function is nonReentrant
             if (pending > 0) {
                 safeTransferReward(msg.sender, pending, _pid);
             }
@@ -59,16 +58,14 @@ contract Deposit is Base {
      * @param _pid The pool id
      */
     function claimReward(uint256 _pid) external {
-        if (!poolInfos[_pid].committeeCheckedIn)
-            revert CommitteeNotCheckedInYet();
         updatePool(_pid);
 
         UserInfo storage user = userInfo[_pid][msg.sender];
         uint256 rewardPerShare = poolInfos[_pid].rewardPerShare;
         if (user.shares > 0) {
             uint256 pending = user.shares * rewardPerShare / 1e12 - user.rewardDebt;
-            user.rewardDebt = user.shares * rewardPerShare / 1e12;
             if (pending > 0) {
+                user.rewardDebt = user.shares * rewardPerShare / 1e12;
                 safeTransferReward(msg.sender, pending, _pid);
             }
         }
--- a/contracts/vaults/Deposit.sol
+++ b/contracts/vaults/Deposit.sol
@@ -13,6 +13,8 @@ contract Deposit is Base {
     * @param _amount Amount of pool's token to deposit. Must be at least `MINIMUM_DEPOSIT`
     **/
     function deposit(uint256 _pid, uint256 _amount) external nonReentrant {
+        if (!poolInfos[_pid].committeeCheckedIn)
+            revert CommitteeNotCheckedInYet();
         if (poolDepositPause[_pid]) revert DepositPaused();
         if (_amount < MINIMUM_DEPOSIT) revert AmountLessThanMinDeposit();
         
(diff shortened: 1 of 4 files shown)


contracts/vaults/Deposit.sol
```diff
@@ -97,6 +96,7 @@ contract Deposit is Base {
         // if the user already has funds in the pool, give the previous reward
         if (_user.shares > 0) {
```
lirona, line 97: we can get rid of passing `_user` , no?

contracts/vaults/Withdraw.sol
```diff
@@ -46,7 +46,7 @@ contract Withdraw is Base {
         updatePool(_pid);
 
         uint256 pending = ((user.shares * pool.rewardPerShare) / 1e12) - user.rewardDebt;
```
lirona, line 48: we should call _claimReward here

test/hatvaults.js
```diff
@@ -1184,13 +1186,55 @@ contract("HatVaults", (accounts) => {
...
+
+    assert.equal(await hatToken.balanceOf(staker), 0);
+
+    let expectedReward = await calculateExpectedReward(staker, 1);
+    await hatVaults.claimReward(0, { from: staker });
+    var tx = await hatVaults.claimReward(0, { from: staker });
+    assert.equal(tx.logs[1].event, "ClaimReward");
+    assert.equal(tx.logs[1].args._pid, 0);
+
+    assert.equal(
```
lirona, line 1234: maybe better to check the balances after the first claim reward, and then check that it hasn't changed after the second one

contracts/vaults/Deposit.sol
ben-kaufman: We can just load user again here but why though?

contracts/vaults/Withdraw.sol
ben-kaufman: Yes but the claim reward has some checks which are unnecessary so not sure because this duplication could save some gas

test/hatvaults.js
ben-kaufman: Well I did it like this because it actually does change, by one block of reward

ben-kaufman: Ok I made it to claim twice in the same block

contracts/vaults/Deposit.sol
jellegerbrandy: let's keep the comments on topic @lirona  :)

contracts/vaults/Deposit.sol
```diff
@@ -97,6 +96,7 @@ contract Deposit is Base {
...
             uint256 pending = _user.shares * poolInfos[_pid].rewardPerShare / 1e12 - _user.rewardDebt;
+            userInfo[_pid][msg.sender].rewardDebt += pending;
```
jellegerbrandy, line 99: this works but it is a very unclear way of writing 

`user.rewardDebt = user.shares * pool.rewardPerShare / 1e12;`

In any case, I guess the update of the rewardDebt should not be part of the `_claimReward` function, bc (see my other comment) it is just useless when we call `_claimReward` as part fo the deposit function

jellegerbrandy (changes requested):

contracts/vaults/Deposit.sol
```diff
@@ -40,7 +40,6 @@ contract Deposit is Base {
 
         user.shares += userShares;
         pool.totalShares += userShares;
-        user.rewardDebt = user.shares * pool.rewardPerShare / 1e12;
```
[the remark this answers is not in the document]
ben-kaufman: Good catch, added a test

test/hatvaults.js
```diff
@@ -1202,6 +1204,46 @@ contract("HatVaults", (accounts) => {
     );
   });
 
+  it("cannot claim reward twice", async () => {
```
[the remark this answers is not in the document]
ben-kaufman: Changed the name

contracts/vaults/Deposit.sol
ben-kaufman: Ok I just removed the _claimReward function and made the calculation clear like before

contracts/vaults/Deposit.sol
```diff
@@ -13,14 +13,23 @@ contract Deposit is Base {
...
         
         //clear withdraw request
         withdrawEnableStartTime[_pid][msg.sender] = 0;
         PoolInfo storage pool = poolInfos[_pid];
         UserInfo storage user = userInfo[_pid][msg.sender];
-        _claimReward(_pid, user);
+        updatePool(_pid);
+        if (user.shares > 0) {
+            uint256 pending = user.shares * pool.rewardPerShare / 1e12 - user.rewardDebt;
+            // no need to update the rewardDebt yet because this function is nonReentrant
```
jellegerbrandy, line 28: this is not a good comment, as the reentrancy can happen by calling another fucntion (like `claimReward` which is not marked nonReentrant)

contracts/vaults/Deposit.sol
```diff
@@ -50,9 +59,19 @@ contract Deposit is Base {
...
+        if (!poolInfos[_pid].committeeCheckedIn)
+            revert CommitteeNotCheckedInYet();
+        updatePool(_pid);
 
-        _claimReward(_pid, user);
+        UserInfo storage user = userInfo[_pid][msg.sender];
+        uint256 rewardPerShare = poolInfos[_pid].rewardPerShare;
+        if (user.shares > 0) {
+            uint256 pending = user.shares * rewardPerShare / 1e12 - user.rewardDebt;
+            user.rewardDebt = user.shares * rewardPerShare / 1e12;
```
jellegerbrandy, line 70: can move this line after `if (pending > 0)` to save some infinitesimal amount of gas :)

jellegerbrandy (changes requested):

contracts/vaults/Deposit.sol
ben-kaufman: Ok so you do want to update rewardDebt here too? Or mark claimReward nonReentrant as well?

jellegerbrandy (approved):

PR #239, opened 2022-06-17
Fix 230

ben-kaufman:


contracts/vaults/Deposit.sol
```diff
@@ -40,7 +41,6 @@ contract Deposit is Base {
 
         user.shares += userShares;
         pool.totalShares += userShares;
-        user.rewardDebt = user.shares * pool.rewardPerShare / 1e12;
```
jellegerbrandy, line 43: let's not mix fixes, can you restore?

contracts/vaults/Deposit.sol
```diff
@@ -97,6 +97,7 @@ contract Deposit is Base {
         // if the user already has funds in the pool, give the previous reward
         if (_user.shares > 0) {
             uint256 pending = _user.shares * poolInfos[_pid].rewardPerShare / 1e12 - _user.rewardDebt;
+            userInfo[_pid][msg.sender].rewardDebt += pending;
```
jellegerbrandy, line 100: pls let's not change this in this pr?

jellegerbrandy (changes requested):

contracts/vaults/Deposit.sol
ben-kaufman: Fixed

contracts/vaults/Deposit.sol
ben-kaufman: Fixed, sorry must have left it there while working on the other stuff

jellegerbrandy (approved):

PR #240, opened 2022-06-17
Fix #234

### LINKED ISSUE #234: arbitrator cannot set bounty (audit #8)
This line needs to be interchanges with the line before:

https://github.com/hats-finance/hats-contracts/blob/e0f2e8026112ba76362192820941ab75858a8baa/contracts/vaults/Claim.sol#L105

(we DEFINITELY need a test for this case)

### LINKED ISSUE #235: validate bountyPercentage in approveClaim (audit #b)
b) Some observations regarding the Claim contract:
- approveClaim._bountyPercentage can go over 100%, but the call will fail with an underflow exception. (assuming 8 is fixed)
- approveClaim._bountyPercentage can surpass bountyInfos[_pid].maxBounty

ben-kaufman:
Fix #234

Fix #235, also fixes some broken tests

### DIFF
--- a/test/arbitrator.js
+++ b/test/arbitrator.js
@@ -1,4 +1,5 @@
 const utils = require("./utils.js");
+const HATTokenLock = artifacts.require("./HATTokenLock.sol");
 const { contract } = require("hardhat");
 const {
   setup,
@@ -80,7 +81,7 @@ contract("HatVaults Arbitrator", (accounts) => {
   });
 
   it("challenge - approve Claim ", async () => {
-    const { hatVaults } = await setup(accounts);
+    const { hatVaults, stakingToken } = await setup(accounts);
     // set challenge period to 1000
     hatVaults.setChallengePeriod(1000);
     const owner = accounts[0];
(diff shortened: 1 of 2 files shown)


test/arbitrator.js
```diff
@@ -100,33 +101,69 @@ contract("HatVaults Arbitrator", (accounts) => {
     const claimId = await submitClaim(hatVaults, { accounts });
 
     // challengeClaim will fail if passing an non-existent claimID
-    assertFunctionRaisesException(
+    await assertFunctionRaisesException(
       hatVaults.challengeClaim("1234", { from: accounts[2] }),
-      "HVE10"
+      "NoActiveClaimExists"
```
jellegerbrandy, line 106: I eam not getting this. Did the test pass before? If so, is the test even working?

ben-kaufman: It was missing `await` in the `assertFunctionRaisesException` call so it wasn't working before, other calls missed it too I think I fixed them all though

jellegerbrandy: Good catch :)

PR #241, opened 2022-06-17
Fix #216

### LINKED ISSUE #216: challengeClaim and dismissClaim should not be timelocked
The task here is to add `challengeClaim` to the list of function that are not timelocked, in the Timelock contract

ben-kaufman:
Fix #216, also fixes wrong error and comment text


jellegerbrandy (approved):
lgtm!

jellegerbrandy (approved):

PR #245, opened 2022-06-20
Fix #233

### LINKED ISSUE #233: arbitrator challenge logic (audit #6)
> Adam Kolar | g0, [09/06/2022 17:57]
> 6) So I don't think the conditions in approveClaim and dismissClaim are quite right. Because:
> - claim can be approved by arbitrator even before challengePeriod expires, but only if it's challenged, so if arbitrator wants to approve claim before challenge period expires, they have to challenge it first. This seems strange.
> - after challengePeriod expires it can be approved by anybody even if it has been challenged, this might be intentional. But I'd expect challenged claims to be only approvable by arbitrator or after challengeTimeOutPeriod expires. With current functionality a dismissalPeriod would be more apt name than challengePeriod.
> - claim can be dismissed by anybody after challengeTimeOutPeriod expires even if it hasn't been challenged. Again this might be intentional.
> - after challengePeriod expires, it can be approved by anybody, but it can also be challenged and dismissed by arbitrator.
> 
> Adam Kolar | g0, [09/06/2022 18:02]
> To put it in different terms, whether or not a claim is challenged currently affects only whether a claim can be dismissed and approved by arbitrator. Otherwise claim being challenged has no effect on the claim.
> 
> Adam Kolar | g0, [09/06/2022 18:04]
> With this in mind and if this functionality is intentional, challengeTimeOutPeriod should be renamed to claimTimeOutPeriod and challengePeriod should be renamed dismissalPeriod.

ben-kaufman:
Fix #233

### DIFF
--- a/contracts/HATTimelockController.sol
+++ b/contracts/HATTimelockController.sol
@@ -26,6 +26,14 @@ contract HATTimelockController is TimelockController {
         hatVaults.approveClaim(_claimId, _bountyPercentage);
     }
 
+    function challengeClaim(uint256 _claimId) external onlyRole(PROPOSER_ROLE) {
+        hatVaults.challengeClaim(_claimId);
+    }
+
+    function dismissClaim(uint256 _claimId) external onlyRole(PROPOSER_ROLE) {
+        hatVaults.dismissClaim(_claimId);
+    }
+
     function addPool(address _lpToken,
                     address _committee,
                     uint256 _maxBounty,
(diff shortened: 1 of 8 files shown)


jellegerbrandy (approved):

jellegerbrandy (changes requested):
These fixes are good, but lets also change the error message in dismissClaim from `OnlyCallableByGovernanceOrAfterChallengeTimeOutPeriod` to `OnlyCallableByArbitratorOrAfterChallengeTimeOutPeriod`

ben-kaufman:
Fixed

jellegerbrandy (dismissed):

jellegerbrandy:
Fix the conflicts and it is good to go :)

jellegerbrandy (approved):

PR #246, opened 2022-06-20
Fix #226 (MINIMUM_DEPOSIT)

### LINKED ISSUE #226: consider removing MINIMAL_DEPOSIT (audit #1)
> 
> 
> Adam Kolar | g0, [09/06/2022 14:46]
> I wonder if instead of having a constant MINIMUM_DEPOSIT it wouldn't be better to have a dynamic minimum deposit value equal to: pool.balance / pool.totalShares
> 
> Adam Kolar | g0, [09/06/2022 14:47]
> If I understand correctly, the purpose of MINIMUM_DEPOSIT is to prevent people from depositing while receiving 0 shares.
> 
> Adam Kolar | g0, [09/06/2022 14:50]
> Note that analogous problem can occur in withdraw after pool balance is reduced by a claim, one share can be worth less than 1 lpToken and that can lead to withdrawal amount being rounded to 0.
> 
> Adam Kolar | g0, [09/06/2022 14:54]
> Also since there's an implication that transferredAmount can differ from _amount (due to transfer fee tokens?) you should probably check minimum deposit against transferredAmount or even better just check that userShares > 0.
> 
> jelle gerbrandy, [09/06/2022 15:41]
> [In reply to Adam Kolar | g0]
> the MINIMUM_DEPOSIT logic is there since v1 and I agree it is a bit clumsy - just checking that userShares > 0 is much more elegant, we should try that out. Personally, I am not so convinced at all that we need special logic to deal with dust.
>

ben-kaufman:
Fix #226 (still need to do same for withdrawal)

### DIFF
--- a/contracts/vaults/Deposit.sol
+++ b/contracts/vaults/Deposit.sol
@@ -67,10 +67,11 @@ contract Deposit is Base {
     function rewardDepositors(uint256 _pid, uint256 _amount) external nonReentrant {
         if ((poolInfos[_pid].balance + _amount) / MINIMUM_DEPOSIT >=
             poolInfos[_pid].totalShares) revert AmountToRewardTooBig();
-
-        uint256 balanceBefore = poolInfos[_pid].lpToken.balanceOf(address(this));
-        poolInfos[_pid].lpToken.safeTransferFrom(msg.sender, address(this), _amount);
-        uint256 lpTokenReceived = poolInfos[_pid].lpToken.balanceOf(address(this)) - balanceBefore;
+        
+        IERC20Upgradeable lpToken = poolInfos[_pid].lpToken;
+        uint256 balanceBefore = lpToken.balanceOf(address(this));
+        lpToken.safeTransferFrom(msg.sender, address(this), _amount);
+        uint256 lpTokenReceived = lpToken.balanceOf(address(this)) - balanceBefore;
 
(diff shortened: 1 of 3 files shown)


contracts/vaults/Deposit.sol
```diff
@@ -58,19 +58,17 @@ contract Deposit is Base {
...
     /**
-     * @notice rewardDepositors - add pool tokens to reward depositors in the pool's native token0
+     * @notice rewardDepositors - add pool tokens to reward depositors in the pool's native token
      * The funds will be given to depositors pro rata upon withdraw
      * The sender of the transaction must have approved the spend before calling this function
      * @param _pid pool id
      * @param _amount amount of pool's native token to add
     */
-    function rewardDepositors(uint256 _pid, uint256 _amount) external nonReentrant {
-        if ((poolInfos[_pid].balance + _amount) / MINIMUM_DEPOSIT >=
```
jellegerbrandy, line 68: Ok, so this was not good after all, see the my comment here: https://github.com/hats-finance/hats-contracts/issues/226#issuecomment-1161530748

jellegerbrandy (changes requested):

jellegerbrandy (changes requested):
ok, instead we decided to remove the `rewardDepositors` function

jellegerbrandy (approved):

PR #275, opened 2022-08-05
Fix #274, HATToken.sol Gas Optimized Now. 

### LINKED ISSUE #274: HATToken.sol not Gas Optimized
Not a bug but it can be a big problem. As the contract HATToken.sol includes so many `require` statements with error messages, it will be too much gas costly. As it stores full-length strings

A simple solution would be to use `Custom Errors` with if-else revert statements, instead

[For More Information, Check This.](https://ethereum.stackexchange.com/questions/101782/requirecondition-message-vs-revert-with-a-custom-error-which-is-better-a)

DevABDee:
HATToken.sol had many `require` statements with Error Messages. It was making it too much gas costly. As error messages were storing full-length strings.
So I replace those `require` statements with the `Custom Errors`. Now the contract is more gas optimized (Alhumduilah)

It's saving **633954 Gas** (this is execution cost and obv the function calling gas is reduced as well)

Before Gas Optimization:
hatsBefore

After Gas Optimization:
hatsAfter

Issue #274 Solved :)
I hope that the Hats Team & the Community will appreciate my small contribution to this beautiful project.
Thanks,
AB Dee.

### DIFF
--- a/contracts/HATToken.sol
+++ b/contracts/HATToken.sol
@@ -4,8 +4,27 @@ pragma solidity 0.8.14;
 import "@openzeppelin/contracts/utils/math/SafeMath.sol";
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 
-contract HATToken is IERC20 {
+error HAT__NotGovernance();
+error HAT__NotGovernancePending();
+error HAT__CannotConfirmGovernanceAtThisTime();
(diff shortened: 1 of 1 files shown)


jellegerbrandy:
Thank you very much @DevABDee for your contribution!

At the moment, we are not actively developing the token contract (even if we are aware of its imperfections) so we are not merging it as of yet..

PR #276, opened 2022-08-08
ERC 4626

ben-kaufman:


contracts/vaults/Claim.sol
```diff
@@ -26,41 +27,34 @@ contract Claim is Base {
...
         // solhint-disable-next-line not-rely-on-time
         if (block.timestamp % (generalParameters.withdrawPeriod + generalParameters.safetyPeriod) <
         generalParameters.withdrawPeriod) revert NotSafetyPeriod();
-        if (_bountyPercentage > bountyInfos[_pid].maxBounty)
+        if (_bountyPercentage > maxBounty)
             revert BountyPercentageHigherThanMaxBounty();
-        uint256 claimId = uint256(keccak256(abi.encodePacked(_pid, block.number, nonce++)));
-        claims[claimId] = Claim({
-            pid: _pid,
+        activeClaim = Claim({
```
jellegerbrandy, line 48: we added a claimId to be able to track the claim in the arbitrator. So please restore that logic.

contracts/HATVaultsRegistry.sol
```diff
@@ -0,0 +1,428 @@
...
+            _rewardController,
+            _descriptionHash,
+            _maxBounty,
+            _bountySplit,
+            _bountyVestingParams[0],
+            _bountyVestingParams[1]
+        );
+    }
+
+    function addTokensToSwap(
```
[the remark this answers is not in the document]
ben-kaufman: Why not? I don't see a risk really, so if someone wants to send tokens to the contract what's the problem?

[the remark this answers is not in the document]
ben-kaufman: Why not? I don't see a risk really, so if someone wants to send tokens to the contract what's the problem?

contracts/HATVaultsRegistry.sol
```diff
@@ -0,0 +1,428 @@
...
+        Each entry is a number between 0 and `HUNDRED_PERCENT`.
+        Total splits should be equal to `HUNDRED_PERCENT`.
+        Bounty must be specified for the hacker (direct or vested in vault's token).
+    * @param _descriptionHash the hash of the vault description.
+    * @param _bountyVestingParams vesting params for the bounty
+    *        _bountyVestingParams[0] - vesting duration
+    *        _bountyVestingParams[1] - vesting periods
+    */
+
+    function createVault(
```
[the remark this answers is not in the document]
ben-kaufman: Well why not have it open now? I mean the vaults aren't connected now so there's no risk really in opening this function

contracts/vaults/Claim.sol
ben-kaufman: But there is only one claim per vault, so you can just as easily use the vault address... it just makes no sense to pass claim ID everywhere if there's only one claim in the contract, it's just extra data with no logic or benefit.

jellegerbrandy: it is because the arbitrator may be async. It is written down somewhere selse, but the scenario roughly is:
- active claim gets disputed (call this claim 1)
- arbittrator takes a long time to decide
- active claim gets resolved bc of arbitrator activity on the vaults contract
- a new claim is made (call this claim 2)
- arbitrator finally decides on result of claim 1
- arbitrator now approves "the active claim", but this is claim 2 and that is Really Wrong

contracts/HATVaultsRegistry.sol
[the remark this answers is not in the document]
jellegerbrandy: Yeah you are right, it is a feature

contracts/HATVaultsRegistry.sol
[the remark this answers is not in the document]
jellegerbrandy: I don't see a risk either, I was just checking this was done on purpose :)

contracts/vaults/Claim.sol
ben-kaufman: But if it's a different claim it will not be challenged, so a delayed call will just fail anyway

contracts/HATVault.sol
```diff
@@ -0,0 +1,93 @@
...
+
+contract HATVault is Claim, Deposits, Params, Withdrawals {
+
+    function _deposit(
+        address caller,
+        address receiver,
+        uint256 assets,
+        uint256 shares
+    ) internal override(Deposits, ERC4626Upgradeable) {
+        Deposits._deposit(caller, receiver, assets, shares);
```
jellegerbrandy, line 20: why copy the function here?

ben-kaufman: It won't compile otherwise. That's another reason I wanted to scrap the multiple files thing and just have it all in one file (also I think the code now is not too long for that anymore).

contracts/HATVault.sol
```diff
@@ -0,0 +1,93 @@
...
+        rewardController.updateVaultBalance(from, amount, false, true);
+    }
+
+    /** @dev See {IERC4626-withdraw}. */
+    function withdraw(
+        uint256 assets,
+        address receiver,
+        address owner
+    ) public override(Withdrawals, ERC4626Upgradeable) returns (uint256) {
+        return Withdrawals.withdraw(assets, receiver, owner);
```
jellegerbrandy, line 48: why the duplicate?

ben-kaufman: https://github.com/hats-finance/hats-contracts/pull/276#discussion_r947907811

contracts/HATVault.sol
```diff
@@ -0,0 +1,93 @@
...
+
+pragma solidity 0.8.14;
+
+
+import "./vaults/Claim.sol";
+import "./vaults/Deposits.sol";
+import "./vaults/Params.sol";
+import "./vaults/Withdrawals.sol";
+
+contract HATVault is Claim, Deposits, Params, Withdrawals {
```
jellegerbrandy, line 12: Can we inherit from some kind of pre-exisitng erc4626 interface?

ben-kaufman: We do https://github.com/hats-finance/hats-contracts/pull/276/files#diff-5ff9b4e6ce18e39f2814e7d6b168516c5ee11e9f3057624fe4f06c33281a6582R82

Again, much nicer as a single file imo

jellegerbrandy: nvm it's in base :)

contracts/HATVault.sol
```diff
@@ -0,0 +1,93 @@
...
+    /** @dev See {IERC4626-maxRedeem}. */
+    function maxRedeem(address owner) public view virtual override returns (uint256) {
+        if (activeClaim.createdAt != 0 || !isWithdrawEnabledForUser(owner)) return 0;
+        return balanceOf(owner);
+    }
+
+    /** @dev See {IERC4626-previewWithdraw}. */
+    function previewWithdraw(uint256 assets) public view virtual override returns (uint256) {
+        uint256 assetsPlusFee = (assets * HUNDRED_PERCENT / (HUNDRED_PERCENT - withdrawalFee));
+       return _convertToShares(assetsPlusFee, MathUpgradeable.Rounding.Up);
```
jellegerbrandy, line 85: indentation

contracts/HATVaultsRegistry.sol
```diff
@@ -0,0 +1,428 @@
...
+    // a struct with parameters for all vaults
+    GeneralParameters public generalParameters;
+    ITokenLockFactory public immutable tokenLockFactory;
+    // feeSetter sets the withdrawal fee
+    address public feeSetter;
+    // address of the arbitrator - which can dispute claims and override the committee's decisions
+    address public arbitrator;
+    // the token into which a part of the the bounty will be swapped-into-and-burnt - this will
+    // typically be HATs
+    ERC20Burnable public immutable swapToken;
```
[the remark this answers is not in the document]
ben-kaufman: Ok updated

contracts/HATVault.sol
ben-kaufman: fixed

contracts/vaults/Base.sol
```diff
@@ -201,171 +126,157 @@ contract Base is Ownable, ReentrancyGuard {
...
+        IRewardController _rewardController,
+        uint256 _vestingDuration,
+        uint256 _vestingPeriods,
+        uint256 _maxBounty,
+        BountySplit memory _bountySplit,
+        IERC20 _asset,
+        address _committee,
+        bool _isPaused
+    ) external initializer {
+        if (_vestingDuration > 120 days)
```
jellegerbrandy, line 237: perhaps use `setVestingParams` here do not duplicate these checks? (i.e. abstract it in an `_setVestingParams internal` and use that)

ben-kaufman: Did that in the internal review PR

ben-kaufman: https://github.com/hats-finance/hats-contracts/pull/281/commits/7b69095b2494db7e376089e5c38836050d5a97d0

contracts/vaults/Base.sol
```diff
@@ -201,171 +126,157 @@ contract Base is Ownable, ReentrancyGuard {
...
+
+    /**
+    * @notice change the information of the vault
+    * ony calleable by the owner of the contract
+    * @param _visible is this vault visible in the UI
+    * @param _depositPause pause deposits (default false).
+    * This parameter can be used by the UI to include or exclude the vault
+    * @param _descriptionHash the hash of the vault's description.
+    */
+    function updateVaultInfo(
```
jellegerbrandy, line 270: move to params.sol

contracts/vaults/Claim.sol
jellegerbrandy: well the only reason it is not challenged in my scenario si bc I forgot to add that step :) - I edit my response

ben-kaufman: But the arbitrator would have to challenge claim 1, then claim 2 in your scenario, which practically acknowledges that there is a new claim.

jellegerbrandy: also, independent of this unlikely (but afaics possible :))  scenario, these claims are objects being passed around in different systems (arbitrator, ui systems) and it will be useful to give them identifiers. Just think of how shitty it is to develop the frontend and yuo have to check the status of a certain claim without an id...

ben-kaufman: If the arbitrator challenges, then challenges again the obviously they know there is a new claim

ben-kaufman: But there is only ever one claim for each vault address. I mean I could add back this claim ID but it still just seems to me like a waste of gas with no added benefit...

contracts/vaults/Withdrawals.sol
```diff
@@ -0,0 +1,142 @@
...
+    function _withdraw(
+        address caller,
+        address receiver,
+        address owner,
+        uint256 assets,
+        uint256 shares,
+        uint256 fee,
+        bool claimReward
+    ) internal nonReentrant {
+        // TODO: If a user gives allowance to another user, that other user can spam to some extent the allowing user's withdraw request
```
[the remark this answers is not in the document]
ben-kaufman: No that's the scenario we already talked and said it's fine. I just forgot to remove the comment.

contracts/vaults/Withdrawals.sol
```diff
@@ -0,0 +1,142 @@
...
+    /** @dev See {IERC4626-redeem}. */
+    function redeem(
+        uint256 shares,
+        address receiver,
+        address owner
+    ) public override virtual returns (uint256) {
+        if (shares > maxRedeem(owner)) revert RedeemMoreThanMax();
+
+        uint256 assets = previewRedeem(shares);
+        uint256 fee = _convertToAssets(shares, MathUpgradeable.Rounding.Down) - assets;
```
jellegerbrandy, line 80: Why not do:
```
uint256 assets = _convertToAssets(shares, MathUpgradeable.Rounding.Down);
uint256 fee = assets * withdrawalFee / HUNDRED_PERCENT;
 ```

jellegerbrandy: is a bit more explicit and also safes a bit of gas

contracts/vaults/Withdrawals.sol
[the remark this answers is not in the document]
jellegerbrandy: cool

contracts/vaults/Claim.sol
jellegerbrandy: > If the arbitrator challenges, then challenges again the obviously they know there is a new claim

This issue came up in the kleros prototype integration here: https://github.com/hats-finance/hats-contracts/pull/133 and so it is not so obvious at all. Kleros will need an ID anyway (I guess any arbitration mechanism would want that) so the question there was more of where it should be generated, and at the vaults seems the most straightforward.

> But there is only ever one claim for each vault address. I 

There is only ever one claim for each vault address **in the vault contract**. But for the arbitrator and for the UI, there will be the entire history of claims. And they will need a way to check which of those claims is "the active claim".

It is possible to resolve these issues in the abitrator integration and in the ui (by creating ids there and keeping careful track of the status in the vaults to know which one of the claims is "the active claim"). But it will be a big pain :)

contracts/vaults/Base.sol
ben-kaufman: I just broke it down here https://github.com/hats-finance/hats-contracts/pull/281/commits/ed47dd31cc4644516bda6fbc58c12511e9ecd94d

contracts/vaults/Claim.sol
ben-kaufman: Ok I added the claim Id: https://github.com/hats-finance/hats-contracts/pull/276/commits/aa1ed0cff73feb374b99d385e126fb01b8b0617e

But it significantly increases gas cost and contract size, while it could be implemented in the UI, so I'm still against having that :)

contracts/vaults/Withdrawals.sol
ben-kaufman: I'm not sure though I think the code is more consistent (with the withdraw function using preview withdraw), as it is now. You think we should change that anyway?

jellegerbrandy (approved):

PR #281, opened 2022-08-16
Internal review fixes

ben-kaufman:


contracts/HATTimelockController.sol
```diff
@@ -29,16 +29,16 @@ contract HATTimelockController is TimelockController {
...
-        _vault.updateVaultInfo(
-            _registered,
-            _depositPause,
-            _descriptionHash
-        );
+    function setDepositPause(HATVault _vault, bool _depositPause) external onlyRole(PROPOSER_ROLE) {
+        _vault.setDepositPause(_depositPause);
+    }
+
+    function updateVaultVisibility(HATVault _vault, bool _visible) external onlyRole(PROPOSER_ROLE) {
```
jellegerbrandy, line 36: lets call this `setVaultVisibility` to be consistent with the other setters

contracts/HATTimelockController.sol
```diff
@@ -29,16 +29,16 @@ contract HATTimelockController is TimelockController {
...
+        _vault.registry().updateVaultVisibility(address(_vault), _visible);
+    }
+
+    function updateVaultDescription(HATVault _vault, string memory _descriptionHash) external onlyRole(PROPOSER_ROLE) {
```
jellegerbrandy, line 40: lets call this `setVaultDescription` to be consistent with the other setters

contracts/RewardController.sol
```diff
@@ -152,28 +157,27 @@ contract RewardController is IRewardController, OwnableUpgradeable {
...
     }
 
-    function updateVaultBalance(
-        address _user,
-        uint256 _sharesChange,
-        bool _isDeposit,
-        bool _claimReward
-    ) external {
-        _updateVaultBalance(msg.sender, _user, _sharesChange, _isDeposit, _claimReward);
+    function updateVaultBalance(address _user, uint256 _sharesChange, bool _isDeposit) external {
```
jellegerbrandy, line 162: feels a bit wrong to have this public, but I see no attack here so np. Perhaps add a docstring saying that this is meant to be called by the vault, but will be harmless if called by a non-vault address

contracts/RewardController.sol
```diff
@@ -152,28 +157,27 @@ contract RewardController is IRewardController, OwnableUpgradeable {
...
+        _updateVaultBalance(msg.sender, _user, _sharesChange, _isDeposit);
     }
 
     /**
      * @notice Transfer to the sender their pending share of rewards.
```
jellegerbrandy, line 167: must be "transfer to _user"

contracts/RewardController.sol
ben-kaufman: Yeah I felt that too at first, but then since there is no harm or risk in having it open I guess it's cleaner and saves gas compared to adding limitations

jellegerbrandy (approved):
lgtm

PR #288, opened 2022-08-22
audit-pre-review

jellegerbrandy:


README.md
```diff
@@ -1 +1,39 @@
-# whi
\ No newline at end of file
+# <img src="https://raw.githubusercontent.com/hats-finance/icons/main/hats.svg" alt="Hats.Finance" text="sds" height="40px"> Hats.Finance | Contracts
+
+[![Coverage Status](https://coveralls.io/repos/github/hats-finance/hats-contracts/badge.svg?t=Ko4Ndz&kill_cache=2)](https://coveralls.io/github/hats-finance/hats-contracts)
+
+The Hats protocol is designed to give white hats hackers the opportunity to gain more on their good behaviour and contribution. Trying to tilt the balance of incentives, and incentivizing more hackers to act responsively. It is doing so by letting projects publish on-chain bounties for their protocols, with committees in-charge of approving or rejecting claims. To further increase the efficiency of this model, the HAT token is introduced, to help bootstrap both ends in this two-sided market.
```
jellegerbrandy, line 5: This is not completely up to date

README.md
```diff
@@ -1 +1,39 @@
...
+
+## Overview
```
jellegerbrandy, line 7: missing doc

README.md
```diff
@@ -1 +1,39 @@
...
+
+Run the tests:
+
+```
+npx hardhat test
+```
+
+## Security
+
+The latest audit was done on August 21, 2021
```
jellegerbrandy, line 27: must update

contracts/HATVault.sol
```diff
@@ -0,0 +1,93 @@
...
+        address from,
+        address to,
+        uint256 amount
+    ) internal virtual override {
+        if (from == address(0) || to == address(0)) {
+            return;
+        }
+        // Users can only deposit for themselves if withdraw request exists
+        if (withdrawEnableStartTime[to] != 0) {
+            revert CannotDepositToAnotherUserWithWithdrawRequest();
```
jellegerbrandy, line 33: not the best error here, should be `CannotTransferToAnotherUserWithWithdrawRequest`

jellegerbrandy: Also this does not cover the case that the withdrawrequest has timed out.

ben-kaufman: Good point

contracts/HATVaultsRegistry.sol
```diff
@@ -0,0 +1,563 @@
...
+        uint256 safetyPeriod;
+        // period of time after withdrawRequestPendingPeriod where it is possible to withdraw
+        // (after which withdrawals are not possible)
+        uint256 withdrawRequestEnablePeriod;
+        // period of time that has to pass after withdraw request until withdraw is possible
+        uint256 withdrawRequestPendingPeriod;
+        // period of time that has to pass after setting a pending max
+        // bounty before it can be set as the new max bounty
+        uint256 setMaxBountyDelay;
+        // fee in ETH to be transferred with every logging of a claim
```
jellegerbrandy, line 88:
```suggestion
        // fee in ETH to be transferred with every logging of a claim using `logClaim`
```

contracts/HATVaultsRegistry.sol
```diff
@@ -0,0 +1,563 @@
...
+        generalParameters.hatVestingPeriods = _periods;
+        emit SetHatVestingParams(_duration, _periods);
+    }
+
+    /**
+    * @notice Called by governance to set the timelock delay for setting the
+    * max bounty (the time between setPendingMaxBounty and setMaxBounty)
+    * @param _delay The time period for the delay. Must be at least 2 days.
+    */
+    function setMaxBountyDelay(uint256 _delay)
```
jellegerbrandy, line 322: do we really need this to be settable?

contracts/HATVaultsRegistry.sol
```diff
@@ -0,0 +1,563 @@
...
+        uint256 hatsReceived;
+        uint256 totalHackerReward;
+    }
+
+    address public immutable hatVaultImplementation;
+    address[] public hatVaults;
+
+    // PARAMETERS FOR ALL VAULTS
+    // time during which a claim can be challenged by the arbitrator
+    uint256 public challengePeriod;
```
jellegerbrandy, line 105: should this not be in generalParameters, just as the other global parameters?

contracts/HATVaultsRegistry.sol
```diff
@@ -0,0 +1,563 @@
...
+
+    struct SwapData {
+        uint256 totalHackersHatReward;
+        uint256 amount;
+        uint256 amountUnused;
+        uint256 hatsReceived;
+        uint256 totalHackerReward;
+    }
+
+    address public immutable hatVaultImplementation;
```
jellegerbrandy, line 100: thinking if we want this to be settable. But probaly no need, we can always get a new registry..

contracts/HATVaultsRegistry.sol
```diff
@@ -0,0 +1,563 @@
...
+    * @return vault The address of the new vault
+    */
+    function createVault(
+        IERC20 _asset,
+        address _committee,
+        IRewardController _rewardController,
+        uint256 _maxBounty,
+        HATVault.BountySplit memory _bountySplit,
+        string memory _descriptionHash,
+        uint256[2] memory _bountyVestingParams,
```
jellegerbrandy, line 357: why is this packed and not just two params?

contracts/HATVaultsRegistry.sol
```diff
@@ -0,0 +1,563 @@
...
+            _maxBounty,
+            _bountySplit,
+            _asset,
+            _committee,
+            _isPaused
+        );
+
+        hatVaults.push(vault);
+
+        emit VaultCreated(
```
jellegerbrandy, line 378: also emit a `SetVaultDescription` event?

contracts/HATVaultsRegistry.sol
```diff
@@ -0,0 +1,563 @@
...
+    }
+
+    /**
+    * @notice change the UI visibility of a vault
+    * only calleable by the owner of the contract
+    * @param _vault the vault to update
+    * @param _visible is this vault visible in the UI
+    * This parameter can be used by the UI to include or exclude the vault
+    */
+    function setVaultVisibility(address _vault, bool _visible) external onlyOwner {
```
jellegerbrandy, line 398: consider storing this explicitly so it can be references by other contracts, #302

contracts/HATVaultsRegistry.sol
```diff
@@ -0,0 +1,563 @@
...
+    */
+    function swapAndSend(
+        address _asset,
+        address[] calldata _beneficiaries,
+        uint256 _amountOutMinimum,
+        address _routingContract,
+        bytes calldata _routingPayload
+    ) external onlyOwner {
+        // Needed to avoid a stack too deep error
+        SwapData memory swapData;
```
jellegerbrandy, line 454: so this is a bit surprising to me as the code is not so complex - you sure of ths @ben-kaufman ?

contracts/HATVaultsRegistry.sol
```diff
@@ -0,0 +1,563 @@
...
+        for (uint256 i = 0; i < _beneficiaries.length; i++) {
+            uint256 hackerReward = swapData.hatsReceived * hackersHatReward[_asset][_beneficiaries[i]] / swapData.amount;
+            swapData.totalHackerReward += hackerReward;
+            hackersHatReward[_asset][_beneficiaries[i]] = swapData.amountUnused * hackersHatReward[_asset][_beneficiaries[i]] / swapData.amount;
+            address tokenLock;
+            if (hackerReward > 0) {
+                // hacker gets her reward via vesting contract
+                tokenLock = tokenLockFactory.createTokenLock(
+                    address(_HAT),
+                    0x000000000000000000000000000000000000dEaD, //this address as owner, so it can do nothing.
```
jellegerbrandy, line 474: wait we still ahve the tokenlock bug to resolve, set this to 0x0000?

contracts/HATVaultsRegistry.sol
ben-kaufman: We now said this will move to be per vault so stored in the vault itself

contracts/HATVaultsRegistry.sol
ben-kaufman: Yes I tried any other way but it gets to stack too deep

contracts/HATVaultsRegistry.sol
```diff
@@ -0,0 +1,563 @@
...
+        address _asset,
+        address[] calldata _beneficiaries,
+        uint256 _amountOutMinimum,
+        address _routingContract,
+        bytes calldata _routingPayload
+    ) external onlyOwner {
+        // Needed to avoid a stack too deep error
+        SwapData memory swapData;
+        for (uint256 i = 0; i < _beneficiaries.length; i++) {
+            swapData.totalHackersHatReward += hackersHatReward[_asset][_beneficiaries[i]];
```
jellegerbrandy, line 456: The caller of this function needs to know the total amount to be swapped to be able to specify a reasonable value for amountOutMininum. This is not so easy now. Should we add a getter for hackersHatReward, so it is easy to read from the outside?

contracts/vaults/Claim.sol
```diff
@@ -0,0 +1,192 @@
...
+        // send to the registry the amount of tokens which should be swapped 
+        // to HAT so it could call swapAndSend in a separate tx.
+        asset.safeApprove(address(registry), claimBounty.hackerHatVested + claimBounty.governanceHat);
+        registry.addTokensToSwap(
+            asset,
+            claim.beneficiary,
+            claimBounty.hackerHatVested,
+            claimBounty.governanceHat
+        );
+
```
jellegerbrandy, line 133: sets do an asset.safeApporve(registyr, 0) here just to be sure

contracts/HATVaultsRegistry.sol
lirona: have you tried putting stuff in a scope? the for loop can be in {} i think

contracts/HATVaultsRegistry.sol
lirona: we have all the info in the approveClaim event so we can read from that. maybe we should have addTokensToSwap emit an event also?

contracts/HATVaultsRegistry.sol
ben-kaufman: Yes I tried any nicer trick but still got the stack too deep issue.

contracts/HATVaultsRegistry.sol
ben-kaufman: That's how it was before, will change that

ben-kaufman: Oh without it I get stack too deep. But maybe if we remove something else like the setting of the description, or the setting of isPaused. What do you think?

contracts/HATVaultsRegistry.sol
lirona: then maybe would be better to specify the minimum ratio instead of the minimum amount and then calculate the minimum amount in the function

contracts/HATVaultsRegistry.sol
lirona: you can also get rid of `totalHackerReward` and calculate the gov reward directly in the end (swapData.hatsReceived * governanceHatReward[_asset] / swapData.amount )

lirona: that's one var, but maybe would be enough? :)

ben-kaufman: Even without these I get stack too deep

contracts/HATVaultsRegistry.sol
ben-kaufman: https://github.com/hats-finance/hats-contracts/pull/306

contracts/HATVaultsRegistry.sol
ben-kaufman: hackersHatReward is public so not sure I understand the issue

contracts/vaults/Claim.sol
ben-kaufman: https://github.com/hats-finance/hats-contracts/pull/307

contracts/HATVaultsRegistry.sol
ben-kaufman: Done

contracts/vaults/Claim.sol
```diff
@@ -0,0 +1,202 @@
...
+        claimBounty.governanceHat = governanceHat;
+        claimBounty.hackerHatVested = hackerHatVested;
+        claimBounty.hackerVested =
+        totalBountyAmount * bountySplit.hackerVested
+        / HUNDRED_PERCENT_SQRD;
+        claimBounty.hacker =
+        totalBountyAmount * bountySplit.hacker
+        / HUNDRED_PERCENT_SQRD;
+        claimBounty.committee =
```
jellegerbrandy, line 198: mb better to just give "the rest" to the comittee so we do not have rounding errors (not that it matters really but it is more precise)

contracts/HATVaultsRegistry.sol
```diff
@@ -0,0 +1,603 @@
...
+    *   Each entry is a number between 0 and `HUNDRED_PERCENT`.
+    *   Total splits should be less than `HUNDRED_PERCENT`.
+    * @param _tokenLockFactory Address of the token lock factory to be used
+    * to create a vesting contract for the approved claim reporter.
+    */
+    constructor(
+        address _hatVaultImplementation,
+        address _hatGovernance,
+        address _HAT,
+        HATBountySplit memory _defaultHATBountySplit,
```
jellegerbrandy, line 191: seems easier UX/more explicit to just have two argument here (`governanceHat` and `hackerHatVested`)

jellegerbrandy: basically I am proposing to remove the HATBountySplit struct altogether..

contracts/vaults/Claim.sol
ben-kaufman: https://github.com/hats-finance/hats-contracts/pull/316

contracts/HATVaultsRegistry.sol
ben-kaufman: Don't have a strong opinion really. Should we do it?

contracts/HATVaultsRegistry.sol
```diff
@@ -0,0 +1,603 @@
...
+    * @param _defaultChallengeTimeOutPeriod The Default challenge timeout
+    */
+    function setDefaultChallengeTimeOutPeriod(uint256 _defaultChallengeTimeOutPeriod) external onlyOwner {
+        validateChallengeTimeOutPeriod(_defaultChallengeTimeOutPeriod);
+        defaultChallengeTimeOutPeriod = _defaultChallengeTimeOutPeriod;
+        emit SetDefaultChallengeTimeOutPeriod(_defaultChallengeTimeOutPeriod);
+    }
+
+    function validateChallengePeriod(uint256 _challengePeriod) public pure {
+        if ( _challengePeriod < 1 days) revert ChallengePeriodTooShort();
```
jellegerbrandy, line 296:
```suggestion
        if (_challengePeriod < 1 days) revert ChallengePeriodTooShort();
```

contracts/HATVault.sol
```diff
@@ -0,0 +1,1084 @@
...
+        address committee;
+        bool isPaused;
+    }
+
+    uint256 public constant NULL_UINT = type(uint256).max;
+    address public constant NULL_ADDRESS = 0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF;
+    uint256 public constant HUNDRED_PERCENT = 10000;
+    uint256 public constant MAX_BOUNTY_LIMIT = 9000; // Max bounty can be up to 90%
+    uint256 public constant HUNDRED_PERCENT_SQRD = 100000000;
+    uint256 public constant MAX_FEE = 200; // Max fee is 2%
```
jellegerbrandy, line 196:
```suggestion
    uint256 public constant MAX_FEE = 200; // Max withdrawal fee is 2%
```
perhaps rename the constant to `MAX_WITHDRAWAL_FEE`

contracts/HATVault.sol
```diff
@@ -0,0 +1,1092 @@
...
+    onlyOwner noActiveClaim noSafetyPeriod {
+        _validateSplit(_bountySplit);
+        bountySplit = _bountySplit;
+        emit SetBountySplit(_bountySplit);
+    }
+
+    /**
+    * @notice Called by the fee setter to set the fee for withdrawals from the
+    * vault.
+    * @param _fee The new fee. Must be smaller then `MAX_FEE`
```
jellegerbrandy, line 572:
```suggestion
    * @param _fee The new fee. Must be smaller or equal than `MAX_FEE`
```

contracts/HATVault.sol
```diff
@@ -0,0 +1,1092 @@
...
+        if (getArbitrator() != msg.sender) revert OnlyArbitrator();
+        _;
+    }
+
+    modifier notEmergencyPaused() {
+        if (registry.isEmergencyPaused()) revert SystemInEmergencyPause();
+        _;
+    }
+
+    modifier noSafetyPeriod() {
```
jellegerbrandy, line 301:
```suggestion
    // will revert if the current block is within the safety period
    modifier noSafetyPeriod() {
```

docs/parameters.md
```diff
@@ -0,0 +1,68 @@
...
+|`isVaultVisible`| false ||per vault|`setVaultVisibility`
+|`owner`| _hatGovernance | || `transferOwnership`, `renounceOwnership` 
+|`arbitrator`|registry.owner| | set per vault | `vault.setArbitrator`
+|`challengePeriod`|3 days | >= 1 days, <= 5 days | set per vault | `vault.setChallengePeriod`
+|`challengeTimeOutPeriod`| 5 weeks | >= 2 days, <= 85 days| set per vault | `vault.setChallengeTimeOutPeriod`
+|`hatBountySplit.governanceHAT` || |set per vault|`vault.setHATBountySplit`
+|`hatBountySplit.hackerHATVested` || |set per vault|`vault.setHATBountySplit`
+
+
+-  `setHATBountySplit` only if there is no active claim and no safetyperiod
```
jellegerbrandy, line 23:
```suggestion
-  `setHATBountySplit` only if there is no active claim
```

docs/parameters.md
```diff
@@ -0,0 +1,68 @@
...
+-  `setArbitrator` only if there is no active claim and no safetyperiod
```
jellegerbrandy, line 24:
```suggestion

```

docs/parameters.md
```diff
@@ -0,0 +1,68 @@
...
+-  `setChallengePeriod` only if there is no active claim and no safetyperiod
```
jellegerbrandy, line 25:
```suggestion

```

docs/parameters.md
```diff
@@ -0,0 +1,68 @@
...
+|`challengePeriod`|3 days | >= 1 days, <= 5 days | set per vault | `vault.setChallengePeriod`
+|`challengeTimeOutPeriod`| 5 weeks | >= 2 days, <= 85 days| set per vault | `vault.setChallengeTimeOutPeriod`
+|`hatBountySplit.governanceHAT` || |set per vault|`vault.setHATBountySplit`
+|`hatBountySplit.hackerHATVested` || |set per vault|`vault.setHATBountySplit`
+
+
+-  `setHATBountySplit` only if there is no active claim and no safetyperiod
+-  `setArbitrator` only if there is no active claim and no safetyperiod
+-  `setChallengePeriod` only if there is no active claim and no safetyperiod 
+-  `setChallengeTimeOutPeriod` only if there is no active claim and no safetyperiod
```
jellegerbrandy, line 26:
```suggestion

```

contracts/HATVault.sol
```diff
@@ -0,0 +1,1092 @@
...
+    }
+
+    /**
+    * @notice Called by the arbitrator to challenge a claim for a bounty
+    * payout that had been previously submitted by the committee.
+    * Can only be called during the challenge period after submission of the
+    * claim.
+    * @param _claimId The claim ID
+    */
+    function challengeClaim(bytes32 _claimId) external onlyArbitrator isActiveClaim(_claimId) {
```
jellegerbrandy, line 407: We should also check here that the claim has not been challenged yet

jellegerbrandy: addressing this here: https://github.com/hats-finance/hats-contracts/pull/317

contracts/HATVault.sol
```diff
@@ -0,0 +1,1092 @@
...
+    }
+
+    modifier noActiveClaim() {
+        if (activeClaim.createdAt != 0) revert ActiveClaimExists();
+        _;
+    }
+
+    modifier isActiveClaim(bytes32 _claimId) {
+        if (activeClaim.createdAt == 0) revert NoActiveClaimExists();
+        if (activeClaim.claimId != _claimId) revert WrongClaimId();
```
jellegerbrandy, line 318: renamed to more specific "ClaimIsNotActive" in #317

contracts/HATVault.sol
```diff
@@ -0,0 +1,1092 @@
...
+    * @param committee The address of the vault's committee 
+    * @param isPaused Whether to initialize the vault with deposits disabled
+    */
+    // Needed to avoid a stack too deep error
+    struct VaultInitParams {
+        IRewardController rewardController;
+        uint256 vestingDuration;
+        uint256 vestingPeriods;
+        uint256 maxBounty;
+        BountySplit bountySplit;
```
jellegerbrandy, line 184: I guess we can avoid nesting params here

contracts/HATVault.sol
```diff
@@ -0,0 +1,1092 @@
...
+    /**
+    * @dev calculate the specific bounty payout distribution, according to the
+    * predefined bounty split and the given bounty percentage
+    * @param _bountyPercentage The percentage of the vault's funds to be paid
+    * out as bounty
+    * @return claimBounty The bounty distribution for this specific claim
+    */
+    function _calcClaimBounty(uint256 _bountyPercentage) internal view returns(ClaimBounty memory claimBounty) {
+        uint256 totalSupply = totalAssets();
+        if (totalSupply == 0) revert VaultBalanceIsZero();
```
jellegerbrandy, line 1050: why throw an error here? I changed this in #317

contracts/HATVault.sol
ben-kaufman: Good catch :)

contracts/RewardController.sol
```diff
@@ -0,0 +1,252 @@
...
+
+    /**
+    * @notice Safe HAT transfer function, transfer rewards from the contract only if there are enough
+    * rewards available.
+    * @param _to The address to transfer the reward to
+    * @param _amount The amount of rewards to transfer
+    * @param _vault The vault address
+   */
+    function safeTransferReward(address _to, uint256 _amount, address _vault) internal {
+        if (rewardToken.balanceOf(address(this)) < _amount)
```
jellegerbrandy, line 94: perhaps we can remove this check, as the transfer call will revert anyway

jellegerbrandy: actually, let's remove this function completely, it is only used in claimReward. And that has its own event. So we can also remove teh SafeTransferREward event.

contracts/RewardController.sol
```diff
@@ -0,0 +1,252 @@
...
+            return;
+        }
+
+        vault.lastRewardBlock = block.number;
+
+        uint256 totalShares = IERC20Upgradeable(_vault).totalSupply();
+
+        if (totalShares != 0) {
+            uint256 reward = getVaultReward(_vault, lastRewardBlock);
+            vault.rewardPerShare += (reward * 1e12 / totalShares);
```
jellegerbrandy, line 123: better have a variable `REWARD_PRECISION = 1e12` and use that everywhere

contracts/RewardController.sol
```diff
@@ -0,0 +1,252 @@
...
+        if (globalVaultsUpdates.length != 0) {
+            vaultInfo[_vault].lastProcessedTotalAllocPoint = globalVaultsUpdates.length - 1;
+        }
+    }
+
+    /**
+     * @notice Called by owner to set reward multipliers
+     * @param _rewardPerEpoch reward multipliers
+    */
+    function setRewardPerEpoch(uint256[24] memory _rewardPerEpoch) external onlyOwner {
```
jellegerbrandy, line 135: so if the rewards has already started, this will treat vaults differently depending on whether they calim their rewards before or after setting this value. 

Need to check how this was handled in v1

contracts/RewardController.sol
```diff
@@ -0,0 +1,252 @@
...
+    ) external initializer {
+        rewardToken = IERC20Upgradeable(_rewardToken);
+        startBlock = _startRewardingBlock;
+        epochLength = _epochLength;
+        rewardPerEpoch = _rewardPerEpoch;
+        _transferOwnership(_hatGovernance);
+    }
+
+    function setAllocPoint(address _vault, uint256 _allocPoint) external onlyOwner {        
+        updateVault(_vault);
```
jellegerbrandy, line 69: so what happens if we call setAllocPoint twice in the same block with different values? In that case, updateVault will not update the rewardPerShares for that vault.

jellegerbrandy: also setAllocPoint can be frontrun with a call to updateVault

jellegerbrandy: this probably does not make a difference, but I am not 100% sure

contracts/RewardController.sol
```diff
@@ -0,0 +1,252 @@
...
+            if (_isDeposit) {
+                userShares += _sharesChange;
+            } else {
+                userShares -= _sharesChange;
+            }
+        }
+        rewardDebt[_vault][_user] = userShares * rewardPerShare / 1e12;
+    }
+
+    function updateVaultBalance(address _user, uint256 _sharesChange, bool _isDeposit) external {
```
jellegerbrandy, line 162: could we renmae this perhaps to "updateUserRewardBalance" or something?

jellegerbrandy: idk

contracts/RewardController.sol
```diff
@@ -0,0 +1,252 @@
...
+
+contract RewardController is IRewardController, OwnableUpgradeable {
+    using SafeERC20Upgradeable for IERC20Upgradeable;
+
+    // Not enough rewards to transfer to user
+    error NotEnoughRewardsToTransferToUser();
+
+    struct VaultInfo {
+        uint256 rewardPerShare;
+        uint256 lastProcessedTotalAllocPoint;
```
jellegerbrandy, line 19: call this `lastProcessedVaultUpdate`, which is more precise?

contracts/RewardController.sol
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+     /**
+    * @notice Calculate rewards for a vault by iterating over the history of totalAllocPoints updates,
+    * and sum up all rewards periods from vault.lastRewardBlock until current block number.
+    * @param _vault The vault address
+    * @param _fromBlock The block from which to start calculation
+    * @return reward
+    */
+    function getVaultReward(address _vault, uint256 _fromBlock) public view returns(uint256 reward) {
```
jellegerbrandy, line 190: seems that this would revert if _fromBlock < startBlock or if fromBlock is from before the lastProcessedTotalAllocpoint

if so, this is apparently not a problem for internal calls, but  is still unexpected behavior that can be handled easily

contracts/RewardController.sol
```diff
@@ -0,0 +1,252 @@
...
+                                                globalVaultsUpdates[i].totalAllocPoint);
+    }
+
+    function getRewardForBlocksRange(uint256 _fromBlock, uint256 _toBlock, uint256 _allocPoint, uint256 _totalAllocPoint)
+    public
+    view
+    returns (uint256 reward) {
+        if (_totalAllocPoint > 0) {
+            uint256 result;
+            uint256 i = (_fromBlock - startBlock) / epochLength + 1;
```
jellegerbrandy, line 214: this will revert if _fromBlock < startBlock, which seems wrong

contracts/RewardController.sol
```diff
@@ -0,0 +1,252 @@
...
+    function initialize(
+        address _rewardToken,
+        address _hatGovernance,
+        uint256 _startRewardingBlock,
+        uint256 _epochLength,
+        uint256[24] memory _rewardPerEpoch
+    ) external initializer {
+        rewardToken = IERC20Upgradeable(_rewardToken);
+        startBlock = _startRewardingBlock;
+        epochLength = _epochLength;
```
jellegerbrandy, line 63: lets require that this is > 0, as we divide by this later and it makes me nervous :)

contracts/RewardController.sol
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    uint256 public constant MULTIPLIERS_LENGTH = 24;
+
+    // Block from which the vaults contract will start rewarding.
+    uint256 public startBlock;
+    uint256 public epochLength;
+    // the ERC20 contract in which rewards are distributed
+    IERC20Upgradeable public rewardToken;
+    // Reward Multipliers
```
jellegerbrandy, line 36:
```suggestion
    // Rewards
```

contracts/RewardController.sol
```diff
@@ -0,0 +1,252 @@
...
+        uint256 lastRewardBlock;
+        uint256 allocPoint;
+    }
+
+    struct VaultUpdate {
+        uint256 blockNumber;// update blocknumber
+        uint256 totalAllocPoint; //totalAllocPoint
+    }
+
+    uint256 public constant MULTIPLIERS_LENGTH = 24;
```
jellegerbrandy, line 29: rename this to NUMBER_OF_EPOCHS

contracts/RewardController.sol
```diff
@@ -0,0 +1,252 @@
...
+
+    // Block from which the vaults contract will start rewarding.
+    uint256 public startBlock;
+    uint256 public epochLength;
+    // the ERC20 contract in which rewards are distributed
+    IERC20Upgradeable public rewardToken;
+    // Reward Multipliers
+    uint256[24] public rewardPerEpoch;
```
jellegerbrandy, line 37: Perhaps rename this to `rewardPerBlock` or `epochToRewardPerBlock`

contracts/RewardController.sol
ben-kaufman: Well it was not settable in V1 so this problem didn't exist. I can't think of a solution that would keep this functionality and not be very complicated.

One thing we can do is allow changing only the epoch that did not yet start.
Another option is to save all the past rewardPerEpochs and times of change and try to calculate based on that, but this is very complex and might require an array that can get too big...
I guess the first option is best but depends on if this functionality will be sufficient

contracts/RewardController.sol
jellegerbrandy, line 37:
```suggestion
    // amount of tokens rewarded in each block, per epoch
    uint256[24] public rewardPerEpoch;
```

contracts/RewardController.sol
ben-kaufman: Maybe _commitUserBalance or something like that?

contracts/RewardController.sol
ben-kaufman: > also setAllocPoint can be frontrun with a call to updateVault

contracts/RewardController.sol
ben-kaufman: Renamed to epochRewardPerBlock

contracts/RewardController.sol
jellegerbrandy: One possible solution:
```
if (block.number < startBlock) { // if rewards have nto started yet, update
   rewardPerEpoch = _rewardPerEpoch;
}  else {

    nextEpoch  = (block.number - startBlock) / epochLength + 1;
     // if rewards are ongoing, update the future rewards but keep past and current
     for (nextEpoch, nextEpoch< TOTAL_NUMBER_OF_EPOCHS, nextEpoch++) {
             rewardPerEpoch[nextEpoch] = _rewardPerEpoch[nextEpoch]
      }
  
} 
```

contracts/RewardController.sol
jellegerbrandy: sounds good

contracts/RewardController.sol
jellegerbrandy: ah that is your first suggestion :)

contracts/RewardController.sol
ben-kaufman: https://github.com/hats-finance/hats-contracts/pull/321

contracts/HATVault.sol
```diff
@@ -0,0 +1,1109 @@
...
+        if (block.timestamp - pendingMaxBounty.timestamp <
+            generalParameters.setMaxBountyDelay)
+            revert DelayPeriodForSettingMaxBountyHadNotPassed();
+        maxBounty = pendingMaxBounty.maxBounty;
+        delete pendingMaxBounty;
+        emit SetMaxBounty(maxBounty);
+    }
+
+    /**
+    * @notice Called by governance to disable all deposits to the vault
```
jellegerbrandy, line 648:
```suggestion
    * @notice Called by owner to disable all deposits to the vault
```

contracts/HATVault.sol
```diff
@@ -0,0 +1,1109 @@
...
+    */
+    function setVaultDescription(string memory _descriptionHash) external onlyOwner {
+        emit SetVaultDescription(_descriptionHash);
+    }
+
+    /**
+    * @notice Called by vault's owner to set the vault's reward controller
+    * @param _newRewardController The new reward controller
+    */
+    function setRewardController(IRewardController _newRewardController) external onlyOwner {
```
jellegerbrandy, line 669: One possible scenario with changing the rewardcontroller:
- the hats rewardcontroller HRC has a reward program for X
- X.rewardController = HRC
- halfway the reward program, X.rewardController changes to another 
- user balances will not be updated anymore
- rewards will be assigned on the basis of the current balance of shares or something (should check, but in any case it goes wrong :)

contracts/HATVault.sol
```diff
@@ -0,0 +1,1109 @@
...
+            if (registry.isEmergencyPaused()) revert SystemInEmergencyPause();
+            // Cannot transfer or mint tokens to a user for which an active withdraw request exists
+            // because then we would need to reset their withdraw request
+            if (withdrawEnableStartTime[to] != 0) {
+                HATVaultsRegistry.GeneralParameters memory generalParameters = registry.getGeneralParameters();
+                // solhint-disable-next-line not-rely-on-time
+                if (block.timestamp < withdrawEnableStartTime[to] + generalParameters.withdrawRequestEnablePeriod)
+                    revert CannotTransferToAnotherUserWithActiveWithdrawRequest();
+            }
+            rewardController.updateVaultBalance(to, amount, true);
```
jellegerbrandy, line 1006: If this returns an error, then all transfers (and deposits and withdrawals) will revert

this is too much power. Can we do a try . catch here?

contracts/HATVault.sol
jellegerbrandy: Other potential problem is setting it to a value that breaks the transfer

contracts/HATVaultsRegistry.sol
```diff
@@ -0,0 +1,603 @@
...
+        // period of time that has to pass after setting a pending max
+        // bounty before it can be set as the new max bounty
+        uint256 setMaxBountyDelay;
+        // fee in ETH to be transferred with every logging of a claim
+        uint256 claimFee;  
+    }
+
+    // How to divide the hats bounties of the vault, in percentages (out of `HUNDRED_PERCENT`)
+    // The precentages are taken from the total bounty
+    struct HATBountySplit {
```
jellegerbrandy, line 96: I suggest we remove this struct, seems it is just adds annoyance in encoding and decoding in the arguments. Make it `BountyGovernanceHATs` and `BountyHackerHATsVested`?

contracts/HATVault.sol
ben-kaufman: Hmmm right, but then what should we do about it? Remove changing the rewardController?

contracts/HATVault.sol
ben-kaufman: Yes, but then we risk the rewardController getting out of sync. Not sure what the best solution would be here

contracts/HATVault.sol
```diff
@@ -0,0 +1,1109 @@
...
+        address caller,
+        address receiver,
+        uint256 assets,
+        uint256 shares
+    ) internal override virtual nonReentrant {
+        if (!committeeCheckedIn)
+            revert CommitteeNotCheckedInYet();
+        if (shares == 0) revert AmountToDepositIsZero();
+        if (withdrawEnableStartTime[receiver] != 0 && receiver == caller) {
+            // clear withdraw request
```
jellegerbrandy, line 960:
```suggestion
            // clear withdraw request if caller deposits in her own account
```

contracts/HATVault.sol
ben-kaufman: Currently it means that whoever can call the setRewardController can freeze the funds which is bad, maybe we should do with try only for withdraw? Or maybe add a delay on changing the reward controller?

ben-kaufman: Or even just remove setRewardController

contracts/HATVault.sol
```diff
@@ -0,0 +1,1109 @@
+// SPDX-License-Identifier: MIT
+// Disclaimer https://github.com/hats-finance/hats-contracts/blob/main/DISCLAIMER.md
```
jellegerbrandy, line 2: note the disclaimer is empty

contracts/HATVault.sol
jellegerbrandy: @ben-kaufman wdyt?

contracts/HATVaultsRegistry.sol
jellegerbrandy: @ben-kaufman wdyt

contracts/HATVault.sol
jellegerbrandy: @ben-kaufman

contracts/HATVault.sol
jellegerbrandy: For (a) the issue of detaching the rewardcontroller while payout is in process: we coudl do an explicit check (say that allocpont must be 0)

For (b) so the callback to the rewardController in vault._beforeTokenTransfer happens before any state changes if this is a simple transfer. Need to check if that is the case for all functions. In that case, that avoids reentrancy attacks. And we can do a try... catch around the updates so that the _beforeTokenTransfer does not revert.

jellegerbrandy: @ben-kaufman

contracts/HATVault.sol
```diff
@@ -0,0 +1,1109 @@
...
+    * The request will be pending for a period of
+    * `HATVaultsRegistry.GeneralParameters.withdrawRequestPendingPeriod`,
+    * after which a withdraw will be possible for a duration of
+    * `HATVaultsRegistry.GeneralParameters.withdrawRequestEnablePeriod`
+    */
+    function withdrawRequest() external nonReentrant {
+        HATVaultsRegistry.GeneralParameters memory generalParameters = registry.getGeneralParameters();
+        // require withdraw to be at least withdrawRequestEnablePeriod+withdrawRequestPendingPeriod
+        // since last withdrawRequest (meaning the last withdraw request had expired)
+        // unless there's been a deposit or withdraw since, in which case withdrawRequest is allowed immediately
```
jellegerbrandy, line 743:
```suggestion

```

contracts/HATVaultsRegistry.sol
```diff
@@ -0,0 +1,603 @@
...
+    }
+
+    /** 
+    * @dev Check that a given hats bounty split is legal, meaning that:
+    *   Each entry is a number between 0 and less than `HUNDRED_PERCENT`.
+    *   Total splits should be less than `HUNDRED_PERCENT`.
+    * function will revert in case the bounty split is not legal.
+    * @param _hatBountySplit The bounty split to check
+    */
+    function validateHATSplit(HATBountySplit memory _hatBountySplit) public pure {
```
jellegerbrandy, line 550: Given that this is governance setting it's own fee, perhaps this should be limited to a maxium?

contracts/HATVaultsRegistry.sol
ben-kaufman: I can remove that :)

contracts/HATVault.sol
```diff
@@ -0,0 +1,770 @@
...
+    }
+
+    function _beforeTokenTransfer(
+        address from,
+        address to,
+        uint256 amount
+    ) internal virtual override {
+        if (to != address(0)) {
+            if (registry.isEmergencyPaused()) revert SystemInEmergencyPause();
+            // Cannot transfer or mint tokens to a user for which an active withdraw request exists
```
lirona, line 656: i was wondering if we can allow that and keep track of the amount the user had when he did the withdraw request (and this way the withdraw would have a limit)

contracts/HATVaultsRegistry.sol
lirona: wdym?

contracts/HATVaultsRegistry.sol
```diff
@@ -0,0 +1,396 @@
...
+            safetyPeriod: 1 hours,
+            setMaxBountyDelay: 2 days,
+            withdrawRequestEnablePeriod: 7 days,
+            withdrawRequestPendingPeriod: 7 days,
+            claimFee: 0
+        });
+
+        defaultArbitrator = _hatGovernance;
+        defaultChallengePeriod = 3 days;
+        defaultChallengeTimeOutPeriod = 5 weeks;
```
lirona, line 115: shouldn't the feeSetter be set here to _hatGovernance ?

contracts/HATVaultsRegistry.sol
lirona: the description hash is already in the event here so i don't see the need

contracts/HATVaultsRegistry.sol
```diff
@@ -0,0 +1,396 @@
...
+        );
+
+        hatVaults.push(vault);
+
+        emit VaultCreated(
+            vault,
+            address(_asset),
+            _committee,
+            _rewardController,
+            _descriptionHash,
```
lirona, line 254: it did bother me that it's not the same order as the func params but that's just ocd i guess :)

contracts/HATVaultsRegistry.sol
lirona: i think anyway to specify minimum ratio makes more sense, no?

contracts/HATVault.sol
```diff
@@ -0,0 +1,770 @@
...
+    */
+    function _deposit(
+        address caller,
+        address receiver,
+        uint256 assets,
+        uint256 shares
+    ) internal override virtual nonReentrant {
+        if (!committeeCheckedIn)
+            revert CommitteeNotCheckedInYet();
+        if (shares == 0) revert AmountToDepositIsZero();
```
lirona, line 616: why check shares and not assets? what's wrong if someone wants to deposit a small amount that doesn't amount to shares?

contracts/HATVaultsRegistry.sol
```diff
@@ -0,0 +1,396 @@
...
+            _maxBounty,
+            _bountySplit,
+            _bountyVestingParams[0],
+            _bountyVestingParams[1]
```
lirona, line 258: _isPaused should be in the event too, right?

contracts/HATVault.sol
```diff
@@ -0,0 +1,770 @@
...
+
+    struct PendingMaxBounty {
+        uint256 maxBounty;
+        uint256 timestamp;
+    }
+
+    uint256 public constant NULL_UINT = type(uint256).max;
+    address public constant NULL_ADDRESS = 0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF;
+    uint256 public constant HUNDRED_PERCENT = 10000;
+    uint256 public constant MAX_BOUNTY_LIMIT = 9000; // Max bounty can be up to 90%
```
lirona, line 73: there are vaults rn that have a 99% bounty, so we can't have this. or have them lower their bounty.. i think 99% limit is good

contracts/HATVaultsRegistry.sol
```diff
@@ -0,0 +1,396 @@
...
+        generalParameters.setMaxBountyDelay = _delay;
+        emit SetMaxBountyDelay(_delay);
+    }
+
+    /** @notice See {IHATVaultsRegistry-createVault}. */
+    function createVault(
+        IERC20 _asset,
+        address _owner,
+        address _committee,
+        IRewardController _rewardController,
```
lirona, line 221: should we have a default reward controller?

contracts/HATVaultsRegistry.sol
```diff
@@ -0,0 +1,396 @@
...
+        uint256 _maxBounty,
```
lirona, line 222: maybe also can have a default?

contracts/HATVault.sol
ben-kaufman: Then they basically give assets and get nothing, they won't be able to withdraw the assets deposited...

contracts/HATVault.sol
```diff
@@ -0,0 +1,816 @@
...
+    }
+
+    /** @notice See {IHATVault-setVaultDescription}. */
+    function setVaultDescription(string memory _descriptionHash) external onlyOwner {
+        emit SetVaultDescription(_descriptionHash);
+    }
+
+    /** @notice See {IHATVault-setRewardController}. */
+    function setRewardController(IRewardController _newRewardController) external onlyRegistryOwner noActiveClaim {
+        rewardController.terminateVaultRewards();
```
jellegerbrandy, line 433: I want to remove this

My reasoning is this. We usually would not to change a rewardcontroller while rewards are in course. And if we do under normal circumstances, we would first stop the reward program, and then change the reward controller.

But if things go wrong, and our rewardcontroller behaves unexpected, we want to be able to change the rewardcontroller without doing a callback to the old one.

contracts/HATVault.sol
jellegerbrandy: I agree this would be better UX (esp bc ppl cannot receive shares if they have a withdrawrequest open, which is just shitty). But it would also add complexity and new edge cases to consider. So idk.

contracts/HATVaultsRegistry.sol
jellegerbrandy: it is easier to index the description by just subscribing to `SetVaultDescription` events, instead of having to subscribe also to the `VaultCreated` event as you have to do now

contracts/HATVaultsRegistry.sol
jellegerbrandy: we talked about that on a call. We have no specific plans but now that createVault is trustless it seems there may be some sense in having an on-chain record of our approval of this particular vault

contracts/HATVaultsRegistry.sol
jellegerbrandy: it does not really matter i think

contracts/HATVaultsRegistry.sol
jellegerbrandy: very minor but valid point, changing this in https://github.com/hats-finance/hats-contracts/pull/340

contracts/HATVaultsRegistry.sol
jellegerbrandy: it is

contracts/HATVault.sol
jellegerbrandy: it is not a matter of opinion :)  We discovered that if you pay out close to 100% of the vault, bad stuff happens with the shares calculation on new deposits. So if we do 99%, then in theory hafter 4 or 5 payouts of 99% we have paid out 99.999999999% of the vault etc. So we set 90% to be save 

This is also a bug in v1, I think, so these 99% vaults are at risk.

contracts/HATVaultsRegistry.sol
jellegerbrandy: why

contracts/HATVaultsRegistry.sol
jellegerbrandy: why

contracts/HATVault.sol
jellegerbrandy: #341

contracts/HATVaultsRegistry.sol
jellegerbrandy: yes

contracts/HATVault.sol
jellegerbrandy: done that here #340

contracts/HATVault.sol
lirona: it was 20 payouts but anyway that's ok :) do we have note of important changes in v2 that we need to inform the projects about?

contracts/HATVault.sol
lirona: why not? the withdraw is done by assets, not shares.. they can withdraw the same amount they would just not get any rewards, no?

contracts/HATVault.sol
lirona: well the case of transfer was not possible at all in v1, and i'm not sure how much it would be in use, if at all.. but doesn't seem that complex to keep amount to withdraw per user and add that to the maxWithdraw check, but sure there's more i don't think of. maybe not worth it

contracts/HATVault.sol
```diff
@@ -0,0 +1,816 @@
...
+                // solhint-disable-next-line not-rely-on-time
+                if (block.timestamp < withdrawEnableStartTime[to] + generalParameters.withdrawRequestEnablePeriod)
+                    revert CannotTransferToAnotherUserWithActiveWithdrawRequest();
+            }
+            rewardController.commitUserBalance(to, amount, true);
+        }
+
+        if (from != address(0)) {
+            if (activeClaim.createdAt != 0) revert ActiveClaimExists();
+            if (!_isWithdrawEnabledForUser(from)) revert InvalidWithdrawRequest();
```
lirona, line 710: seems to me this is redundant, because this check is done on maxRedeem/maxWithdraw

lirona: or we can remove this check from `withdraw` and `redeem`

contracts/HATVaultsRegistry.sol
lirona: then it defaults to address(0) and you have to call `setFeeSetter`

contracts/HATVaultsRegistry.sol
lirona: because most vault will probably use the default reward plan, no?

contracts/HATVaultsRegistry.sol
lirona: also seems to me like our MAX_BOUNTY_LIMIT would be enough for most projects

contracts/HATVaultsRegistry.sol
lirona: ok, and it's useful? given that it's just the hash and not the description?

contracts/HATVaultsRegistry.sol
lirona: ok. when we talked about it i thought we're also going to use it

contracts/RewardController.sol
```diff
@@ -0,0 +1,272 @@
...
+    function commitUserBalance(address _user, uint256 _sharesChange, bool _isDeposit) external {
+        _commitUserBalance(msg.sender, _user, _sharesChange, _isDeposit);
+    }
+
+    /**
+    * @notice Transfer to the specified user their pending share of rewards.
+    * @param _vault The vault address
+    * @param _user The user address to claim for
+    */
+    function claimReward(address _vault, address _user) external {
```
lirona, line 178: so anyone can call this for anyone now.. i thought anyway this should be in the vault and not the reward controller so maybe better to have it there and call this function from there? like this we will get both values from msg.sender

jellegerbrandy: it not just bad but also good that anyone can call this for anyone. It makes intergrations easier (someone can claim your rewards for you as a sservice)

contracts/HATVault.sol
ben-kaufman: We need it in maxWithdraw and maxRedeem to comply with ERC4626 specs, and here to make sure it is checked also on token transfers.

ben-kaufman: Or actually we could do something like this: https://github.com/hats-finance/hats-contracts/pull/347

contracts/HATVaultsRegistry.sol
jellegerbrandy: > hackersHatReward is public so not sure I understand the issue
nevermind you right


> i think anyway to specify minimum ratio makes more sense, no?

No it does not. The price depends on the amount (if you sell 1 ether for 2000/ETH does not mean you can sell 1B ether for 2000/ETH). So setting the mimim price we want to get for the entire batch is the logical thing. Specifying the average price of the batch is not better UX.

contracts/HATVault.sol
```diff
@@ -0,0 +1,828 @@
...
+                IHATVaultsRegistry.GeneralParameters memory generalParameters = registry.getGeneralParameters();
+                // solhint-disable-next-line not-rely-on-time
+                if (block.timestamp < withdrawEnableStartTime[to] + generalParameters.withdrawRequestEnablePeriod)
+                    revert CannotTransferToAnotherUserWithActiveWithdrawRequest();
+            }
+            rewardController.commitUserBalance(to, amount, true);
+        }
+        // withdraw/redeem/transfer
+        if (from != address(0)) {
+            if (amount > maxRedeem(from)) revert RedeemMoreThanMax();
```
jellegerbrandy, line 720: So one drawback of this pattern is that users will  get error messages of the type "RedeemMoreThanMax" also when more pecific conditions apply , like "you have no open withdraw request" or "you are in safety period".

contracts/RewardController.sol
lirona: i don't see the use case but ok :) and don't you think it should be on the vault? since withdrawAndClaim and redeemAndClaim are there?

contracts/RewardController.sol
```diff
@@ -0,0 +1,276 @@
...
+                rewardDebt[_vault][_user];
+    }
+
+    /**
+    * @notice Transfer any tokens held in this contract to the owner
+    * @param _token the token to sweep
+    * @param _amount the amount of token to sweep
+    */
+    function sweepToken(IERC20Upgradeable _token, uint256 _amount) external onlyOwner {
+        _token.safeTransfer(msg.sender, _amount);
```
lirona, line 269: why to pass the token and not use the `rewardToken` ?

contracts/RewardController.sol
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    uint256 public constant REWARD_PRECISION = 1e12;
+    uint256 public constant NUMBER_OF_EPOCHS = 24;
+
+    // Block from which the vaults contract will start rewarding.
+    uint256 public startBlock;
+    uint256 public epochLength;
+    // the ERC20 contract in which rewards are distributed
+    IERC20Upgradeable public rewardToken;
```
lirona, line 39: why does it have to be upgradeable?

contracts/RewardController.sol
```diff
@@ -0,0 +1,276 @@
...
+    /**
+    * @notice Transfer any tokens held in this contract to the owner
+    * @param _token the token to sweep
+    * @param _amount the amount of token to sweep
+    */
+    function sweepToken(IERC20Upgradeable _token, uint256 _amount) external onlyOwner {
+        _token.safeTransfer(msg.sender, _amount);
+    }
+
+    function getGlobalVaultsUpdatesLength() external view returns (uint256) {
```
lirona, line 272: wanted to remove this for the same reason, it's used for tests

contracts/RewardController.sol
jellegerbrandy: this is the right place because all the relevant information is stored in this contract. Also, in this way users can claim their rewards also after we cahnage the rewardcontroller on the vault.

contracts/RewardController.sol
jellegerbrandy: perhaps other tokens were sent to the contract by mistake

jellegerbrandy: so it covers that one more use case for free

contracts/RewardController.sol
jellegerbrandy: It  does not really matter, but perhaps using IERC20 is slightly more elegant

contracts/RewardController.sol
```diff
@@ -0,0 +1,276 @@
...
+                                            globalVaultsUpdates[i].totalAllocPoint);
+            _fromBlock = nextUpdateBlock;
+        }
+        return reward + getRewardForBlocksRange(_fromBlock,
+                                                block.number,
+                                                vaultAllocPoint,
+                                                globalVaultsUpdates[i].totalAllocPoint);
+    }
+
+    function getRewardForBlocksRange(uint256 _fromBlock, uint256 _toBlock, uint256 _allocPoint, uint256 _totalAllocPoint)
```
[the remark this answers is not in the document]
jellegerbrandy: So making it public almost costs nothing, so I guess there is no compelling reason to make it internal.

WE are also _testing_ the function in the tests, so that is actually good :)

contracts/tokenlock/TokenLock.sol
```diff
@@ -0,0 +1,393 @@
...
+ *
+ * The contract supports receiving extra funds than the managed tokens ones that can be
+ * withdrawn by the beneficiary at any time.
+ *
+ * A releaseStartTime parameter is included to override the default release schedule and
+ * perform the first release on the configured time. After that it will continue with the
+ * default schedule.
+ */
+// solhint-disable-next-line indent
+abstract contract TokenLock is OwnableInitializable, ITokenLock {
```
lirona, line 32: why don't we use openzeppelin tockenLock?

jellegerbrandy: we inherited this and we have no reason to change to OZ now

contracts/HATVaultsRegistry.sol
lirona: ok gotcha :) i actually thought i does mean that.....

contracts/RewardController.sol
lirona: i think so

contracts/RewardController.sol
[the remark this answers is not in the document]
lirona: ok but i didn't include this and `getGlobalVaultsUpdatesLength` in the interface, makes sence?

contracts/RewardController.sol
lirona: actually i tried it and got `Error: Contract `RewardController` is not upgrade safe` .. so nevermind :)

contracts/HATVault.sol
ben-kaufman: Yes but I thought it's worth it compared to the gas saving, the rest can be analyzed easily in the UI

PR #292, opened 2022-08-22
Remove committee check in

### LINKED ISSUE #284: Remove committee check in
The committee check in just pauses vault deposits until the committee starts them for the first time. This seems practically the same as having a 0 address check, which is quite unnecessary. Also, note that we could just start vaults with depositPaused as true until governance starts them to give some time to do checks before starting the vault. Also, not that currently, the committee check in is only for the first time the committee is set, but if after check in the committee is replaced, there is no need committee check in which seems inconsistent.

ben-kaufman:
Fix #284

### DIFF
--- a/contracts/vaults/Params.sol
+++ b/contracts/vaults/Params.sol
@@ -6,20 +6,10 @@ import "./Base.sol";
 contract Params is Base {
 
     /**
-    * @notice Set new committee address. Can be called by existing committee if it had checked in, or
-    * by the governance otherwise.
+    * @notice Set new committee address. Can be called only by the existing committee.
     * @param _committee new committee address
     */
-    function setCommittee(address _committee)
-    external {
-        // governance can update committee only if committee was not checked in yet.
-        if (msg.sender == registry.owner() && committee != msg.sender) {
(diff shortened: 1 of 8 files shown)


contracts/vaults/Params.sol
```diff
@@ -6,20 +6,10 @@ import "./Base.sol";
 contract Params is Base {
 
     /**
-    * @notice Set new committee address. Can be called by existing committee if it had checked in, or
-    * by the governance otherwise.
+    * @notice Set new committee address. Can be called only by the existing committee.
```
lirona, line 9:
```suggestion
    * @notice Called by the existing committee to set a new committee address.
```

lirona: just for consistency, i wrote everything that way

lirona (commented):
lgtm, accept if i understood correctly, we said we also always initialize the vaults with deposits paused, and have committee also be able to change paused status

jellegerbrandy:
I am closing this - in the end it seems better to keep the committee checkin to preserve non-rugpull safety :)

PR #293, opened 2022-08-22
Swap burn send updates

### LINKED ISSUE #278: Funds might get stuck if on swap the routing contract does not use the full amount
If for any reason the routing contract does not use the full amount of the asset to swap for hats, the rest will stay in the contract, but not be accounted for in the balance still available to swap, this will make it stuck in the contract. Should we handle this case (considering that normally the swap contract is expected to use the full amount available) ?

ben-kaufman:
- Remove burn of swapped tokens
- Update amounts used correctly when router only swaps partial amount ( Fix #278 )

### DIFF
--- a/contracts/HATVaultsRegistry.sol
+++ b/contracts/HATVaultsRegistry.sol
@@ -5,7 +5,6 @@ pragma solidity 0.8.16;
 
 import "@openzeppelin/contracts/access/Ownable.sol";
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
-import "@openzeppelin/contracts/token/ERC20/extensions/ERC20Burnable.sol";
 import "@openzeppelin/contracts/proxy/Clones.sol";
 import "./tokenlock/TokenLockFactory.sol";
 import "./interfaces/IRewardController.sol";
@@ -57,7 +56,6 @@ error ChallengeTimeOutPeriodTooLong();
 /// https://github.com/hats-finance/hats-contracts
 contract HATVaultsRegistry is Ownable {
     using SafeERC20 for IERC20;
-    using SafeERC20 for ERC20Burnable;
 
     struct GeneralParameters {
         uint256 hatVestingDuration;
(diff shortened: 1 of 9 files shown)


jellegerbrandy (approved):
looks good to me

contracts/HATVaultsRegistry.sol
```diff
@@ -79,9 +77,8 @@ contract HATVaultsRegistry is Ownable {
     struct SwapData {
```
lirona, line 77: i know it's too late but i don't get why to create this struct if all it's members are just internal function variables and not passed anywhere.. makes more sense to me to just have them as local variables in the one function where you use it.. but it's not important so feel free to ignore :)

docs/roles.md
```diff
@@ -27,7 +27,7 @@ The `owner` of a hatvault created by the registry is the same registry owner.
 - can call `setVaultDescription` to change the vault's description hash
 - can call `setDepositPause` to pause and unpause deposits to the vault
 - can call `approveClaim` to approve a claim for a bounty payout that was previously submitted by a vault's committee
-- can call `swapBurnSend` and swap, burn, and send the vauit's token that are earmarked for payout (after approveClaim)
+- can call `swapAndSend` and swap and send the vauit's token that are earmarked for payout (after approveClaim)
```
lirona, line 30:
```suggestion
```

lirona: it's only called by the registry owner

lirona (commented):
lgtm

contracts/HATVaultsRegistry.sol
ben-kaufman: Because this allows to overcome the stack too deep error

jellegerbrandy: there's a comment in the code that is it to avoid stack too deep erros

lirona: ohhh ok missed that sorry

PR #300, opened 2022-08-24
Allow deposit to user with expired withdraw request

ben-kaufman:
Allow transfer and deposit to user with expired withdraw request

### DIFF
--- a/contracts/HATVault.sol
+++ b/contracts/HATVault.sol
@@ -130,21 +130,21 @@ contract HATVault is Claim, Deposits, Params, Withdrawals {
         address to,
         uint256 amount
     ) internal virtual override {
-        if (from == address(0) || to == address(0)) {
-            return;
+        if (to != address(0)) {
+            // Users can only deposit for themselves if an active withdraw request exists
+            if (withdrawEnableStartTime[to] != 0) {
+                HATVaultsRegistry.GeneralParameters memory generalParameters = registry.getGeneralParameters();
+                // solhint-disable-next-line not-rely-on-time
+                if (block.timestamp < withdrawEnableStartTime[to] + generalParameters.withdrawRequestEnablePeriod)
+                    revert CannotTransferToAnotherUserWithActiveWithdrawRequest();
+            }
+            rewardController.updateVaultBalance(to, amount, true);
         }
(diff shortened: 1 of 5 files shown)


contracts/vaults/Deposits.sol
```diff
@@ -27,7 +27,10 @@ contract Deposits is Base {
         if (shares == 0) revert AmountToDepositIsZero();
         // Users can only deposit for themselves if withdraw request exists
         if (withdrawEnableStartTime[receiver] != 0 && receiver != caller) {
-            revert CannotDepositToAnotherUserWithWithdrawRequest();
+            HATVaultsRegistry.GeneralParameters memory generalParameters = registry.getGeneralParameters();
```
jellegerbrandy, line 30: so:

a) seems to me this check is not needed, as `super._deposit(.. receiver..)` will call `_beforeTokenTransfer` and do the check

b) are you sure users can deposit for themselves if a withdraw request exists? Seems to me it would fail on the new _beforeTokenTransfer (I could not find a test, but it is hard to find tests :)

REcommendation :) remove the lines here in deposit and handle all the logic in _beforeTokentransfer

jellegerbrandy (changes requested):

contracts/vaults/Deposits.sol
ben-kaufman: The way I implemented the _beforeTokentransfer is that it is ignored for both mint and burn operations so it does not apply here. You think I should change that?

contracts/HATVault.sol
```diff
@@ -135,7 +135,10 @@ contract HATVault is Claim, Deposits, Params, Withdrawals {
         }
         // Users can only deposit for themselves if withdraw request exists
```
jellegerbrandy, line 136:
```suggestion
        
```

contracts/vaults/Deposits.sol
jellegerbrandy: ok I see. It seems cleaner to do it in one place.

jellegerbrandy (dismissed):
lgtm

jellegerbrandy (approved):

lirona (approved):

PR #308, opened 2022-08-26
Allow vaults to have values stick to registry defaults

ben-kaufman:


contracts/vaults/Base.sol
```diff
@@ -222,6 +253,54 @@ contract Base is ERC4626Upgradeable, OwnableUpgradeable, ReentrancyGuardUpgradea
...
+            return arbitrationConfig.arbitrator;
+        } else {
+            return registry.defaultArbitrator();
+        }
+    }
+
+    /** 
+    * @notice Returns the vault challenge period
+    */
+    function getChallengePeriod() public view returns(uint256) {
```
jellegerbrandy, line 285: is it not nicer to do something like 
```
function getChallengePeriod() { return arbitrationconfig.challengePeriod || registry.getChallengePeriod())
```

ben-kaufman: How would that work?

contracts/HATVaultsRegistry.sol
```diff
@@ -121,7 +129,11 @@ contract HATVaultsRegistry is Ownable {
     // asset => amount
     mapping(address => uint256) public governanceHatReward;
 
-    HATBountySplit public hatBountySplit;
+    HATBountySplit public defaultHATBountySplit;
+
+    address public defaultArbitrator;
```
jellegerbrandy, line 134: afaic these vars can also be just alled `arbitrator`, `challengePeriod`, etc.

ben-kaufman: You don't think it's clearer/ more accurate like this?

contracts/HATVault.sol
```diff
@@ -91,16 +91,21 @@ contract HATVault is Claim, Deposits, Params, Withdrawals {
...
         _transferOwnership(_params.owner);
         tokenLockFactory = _registry.tokenLockFactory();
-        arbitrator = _registry.owner();
-        challengePeriod = 3 days;
-        challengeTimeOutPeriod = 5 weeks;
+
+        // Set vault to use default registry values where applicable
+        hatBountySplit = HATVaultsRegistry.HATBountySplit({
+            governanceHat: NULL_UINT,
+            hackerHatVested: 0
```
jellegerbrandy, line 104: no NULL_UINT here as well?

contracts/HATVaultsRegistry.sol
jellegerbrandy: it can also be more confusing in some cases. As in "so this is the defaultArbitrator, where is the actual arbitrator defined?"

jellegerbrandy: it is more like a fallback value

contracts/HATVault.sol
ben-kaufman: Well it just saves gas but gets the same result. Since it's a struct I only check the governanceHat, so making this 0 is cheaper

contracts/HATVaultsRegistry.sol
jellegerbrandy: "default" is a bit ambiguous, which is why we had such different interpretations of how this should be implemented

ben-kaufman: I disagree but not very important tbh. So how you call it? Also rename the event and function name?

contracts/vaults/Base.sol
```diff
@@ -160,14 +152,18 @@ contract Base is ERC4626Upgradeable, OwnableUpgradeable, ReentrancyGuardUpgradea
...
     bool public depositPause;
 
+    mapping(address => uint256) public withdrawEnableStartTime;
+
+    // the HATBountySplit of the vault
+    HATVaultsRegistry.HATBountySplit internal hatBountySplit;
+
     // address of the arbitrator - which can dispute claims and override the committee's decisions
-    address public arbitrator;
+    address internal arbitrator;
```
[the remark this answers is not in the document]
jellegerbrandy: ah ok nvm :)

docs/roles.md
```diff
@@ -5,7 +5,22 @@ The contracts in this repository define a number of different roles that have th
 
 ## `HATVaultsRegistry.owner`
 
-- `swapAndSend` and swap and send the vauit's token that are earmarked for payout (after approveClaim)
+- can call `setFeeSetter` set the feeSetter role
```
jellegerbrandy, line 8: Can you undo these changes?

jellegerbrandy: this is a merge going wrong, we can update later

contracts/HATVaultsRegistry.sol
jellegerbrandy: let's just leave this for now and merge as is

jellegerbrandy (approved):

docs/roles.md
ben-kaufman: Ok yeah something went wrong with the conflicts, I just made it as it is now

PR #322, opened 2022-08-28
Allow changing the reward per block only for future epochs

ben-kaufman:


contracts/RewardController.sol
```diff
@@ -112,10 +112,20 @@ contract RewardController is IRewardController, OwnableUpgradeable {
 
     /**
      * @notice Called by owner to set reward per epoch
-     * @param _epochRewardPerBlock reward mper epoch
+     * Reward can only be set for epochs which have not yet started
+     * @param _epochRewardPerBlock reward per block for each epoch
     */
     function setEpochRewardPerBlock(uint256[24] memory _epochRewardPerBlock) external onlyOwner {
-        epochRewardPerBlock = _epochRewardPerBlock;
+        // if rewards have nto started yet, update the full list
```
jellegerbrandy, line 119:
```suggestion
        // if rewards have not started yet, update the full list
```

contracts/RewardController.sol
```diff
@@ -112,10 +112,20 @@ contract RewardController is IRewardController, OwnableUpgradeable {
...
+        if (block.number < startBlock) {
+            epochRewardPerBlock = _epochRewardPerBlock;
+        }  else {
+            uint256 nextEpoch = (block.number - startBlock) / epochLength + 1;
+            // if rewards are ongoing, update the future rewards but keep past and current
+            for (; nextEpoch < NUMBER_OF_EPOCHS; nextEpoch++) {
+                epochRewardPerBlock[nextEpoch] = _epochRewardPerBlock[nextEpoch];
+            }
+        }
         emit SetEpochRewardPerBlock(_epochRewardPerBlock);
```
jellegerbrandy, line 129:
```suggestion
        emit SetEpochRewardPerBlock(epochRewardPerBlock);
```
seems better to return the actual updated state

jellegerbrandy (changes requested):
one small suggestion for change

jellegerbrandy (approved):

PR #325, opened 2022-08-29
Hat vault fixes

### LINKED ISSUE #328: save aribtrator info with the claim (and with the times etc)

### LINKED ISSUE #329: add a flag: aribtrator can just say yes-no or change the bounty size at will. The flag should work like the other arbitrator settings: settable by registry.owner, there is a default seting and a per-vault override

### LINKED ISSUE #331: hatsbountysplit maximum is 20%

### LINKED ISSUE #332: governance can directly challenge a claim

ben-kaufman:
fix #331
fix #332 
fix #328 
fix #329


contracts/HATVault.sol
```diff
@@ -672,18 +674,37 @@ contract HATVault is ERC4626Upgradeable, OwnableUpgradeable, ReentrancyGuardUpgr
...
-    * an approval. Either sets it to the value passed, or to the special "null" vaule, 
+    * @notice Called by governance to set the percentage of each claim bounty that
+    * will be swapped for hats and sent to the governance.
+    * Either sets it to the value passed, or to the special "null" vaule, 
+    * making it always use the registry's default value.
+    * @param _bountyGovernanceHAT The HAT bounty for governance
+    */
+    function setBountyGovernanceHAT(uint256 _bountyGovernanceHAT) external onlyRegistryOwner {
+        if (_bountyGovernanceHAT != NULL_UINT) {
+            registry.validateHATSplit(_bountyGovernanceHAT, getBountyHackerHATVested());
```
jellegerbrandy, line 685: I think it is beter to have a setter in which they are set together, given that also the validation is separate.

so `setBountyHat(_bountyGoverannceHAt, _bountyHackerHATVested)`

jellegerbrandy (dismissed):
lgtm

just the merge conflicts :)

ben-kaufman:
Btw it would be nice too have more test cases for https://github.com/hats-finance/hats-contracts/issues/328
For now I just have a simple one that doesn't cover everything, I think it's fine for now and we should make an issue to add more tests when we have time

contracts/HATVault.sol
```diff
@@ -237,30 +243,31 @@ contract HATVault is IHATVault, ERC4626Upgradeable, OwnableUpgradeable, Reentran
...
             ) revert UnchallengedClaimCanOnlyBeApprovedAfterChallengePeriod();
         }
 
         address tokenLock;
 
-        IHATVault.ClaimBounty memory claimBounty = _calcClaimBounty(claim.bountyPercentage);
+        IHATVault.ClaimBounty memory claimBounty = _calcClaimBounty(
+            claim.bountyPercentage,
+            claim.bountyGovernanceHAT,
+            claim.bountyHackerHATVested
```
jellegerbrandy, line 269: I guess this should also take account the `claim.hacker` and `claim.hackerVested` and `claim.committee` of the current claim instead of using the settings, right?

Perhaps the signature could be `_calcClaimBounty(claim)`

ben-kaufman: Well these can't change during claim so no reason

ben-kaufman: setBountySplit is protected with noActiveClaim

PR #337, opened 2022-08-29
swapandsend event for gov swap

lirona:


jellegerbrandy (commented):
cool! it needs a test though :)

lirona:
> cool! it needs a test though :)

added one :)

PR #340, opened 2022-08-30
even more tweaks

### LINKED ISSUE #339: update openzeppelin deps
https://github.com/hats-finance/hats-contracts/security/dependabot

Let's update packages to latest versions where we can.

jellegerbrandy:
fix #339


contracts/HATVault.sol
```diff
@@ -663,6 +662,7 @@ contract HATVault is IHATVault, ERC4626Upgradeable, OwnableUpgradeable, Reentran
         super._deposit(caller, receiver, assets, shares);
     }
 
+    // amount of shares correspond with assets + fee
```
lirona, line 665: maybe should be mentioned in the public docs as well? i can add

lirona (approved):
thanks for considering my ocd :)

PR #345, opened 2022-08-30
Remove terminateVaultRewards

ben-kaufman:


contracts/RewardController.sol
```diff
@@ -241,6 +241,9 @@ contract RewardController is IRewardController, OwnableUpgradeable {
     * @param _user the account for which the reward is calculated
     */
     function getPendingReward(address _vault, address _user) external view returns (uint256) {
+        if (IHATVault(_vault).rewardControllerRemoved(address(this))) {
+            return 0;
```
jellegerbrandy, line 245: I guess thsis hould be `return unclaimedReward[_vault][_user]`

ben-kaufman: No because it also calculates the uncommitted reward to be claimed

ben-kaufman: Oh right yes

jellegerbrandy (approved):

PR #350, opened 2022-08-31
reward controller interface

lirona:


lirona:
so this time test fails agaim.. on my computer it runs fine.. should i just remove it? or change it to be a bit more flexible? (now it accepts 252 seconds but sometime it comes out as 253

jellegerbrandy (approved):
ok great :)

PR #359, opened 2022-09-18
Adam Audit fixes

ben-kaufman:
https://github.com/hats-finance/hats-contracts/issues/365

adam's comments:
1) this check: https://github.com/hats-finance/hats-contracts/blob/1272ef58df5d65fcc8d4cddbb23772cd4a6a5cf3/contracts/HATVault.sol#L498 can be bypassed by doing a zero token transfer, which will always succeed and will reset withdrawEnableStartTime[from] to 0
see:
https://github.com/hats-finance/hats-contracts/blob/1272ef58df5d65fcc8d4cddbb23772cd4a6a5cf3/contracts/HATVault.sol#L725
-> suggests to remove the check in https://github.com/hats-finance/hats-contracts/blob/1272ef58df5d65fcc8d4cddbb23772cd4a6a5cf3/contracts/HATVault.sol#L498

2) these two intervals:
https://github.com/hats-finance/hats-contracts/blob/1272ef58df5d65fcc8d4cddbb23772cd4a6a5cf3/contracts/HATVault.sol#L714
and
https://github.com/hats-finance/hats-contracts/blob/1272ef58df5d65fcc8d4cddbb23772cd4a6a5cf3/contracts/HATVault.sol#L763

overlap, meaning if the block timestamp is just right, you can receive shares and send them in the same block. Not sure if it's exploitable, but should be fixed.

-> change `<=` to `<` in https://github.com/hats-finance/hats-contracts/blob/1272ef58df5d65fcc8d4cddbb23772cd4a6a5cf3/contracts/HATVault.sol#L763

3) transfer of shares with the same to and from will lead to weird behavior due to the commitUserBalance calls, most likely ending in an underflow error, but might be even exploitable if rounding issues are exploited

-> fixed by fixing point 2

4) A user can call HATVault.transferFrom with 0 amount and arbitrary from which will always pass. This will lead to withdrawEnableStartTime[from] being set to 0. This basically allows anybody to block all transactions of any address.

-> add check in `_beforeTokenTransfer`

5) HATVault.withdraw is confusing and I'm not entierely sure whether it's correct. Especially rounding up when converting to assets here seems dubious. https://github.com/hats-finance/hats-contracts/blob/1272ef58df5d65fcc8d4cddbb23772cd4a6a5cf3/contracts/HATVault.sol#L532
Instead of the back and forth conversion, I'd recommend a version of calculating fee that's more easy to reason about.

    /** @notice See {IHATVault-withdraw}. */ 
    function withdraw(uint256 assets, address receiver, address owner)  
        public override(IHATVault, ERC4626Upgradeable) virtual returns (uint256) { 
        uint256 shares = previewWithdraw(assets); 
        uint256 fee = assets * HUNDRED_PERCENT / (HUNDRED_PERCENT - withdrawalFee) - assets; 
        _withdraw(_msgSender(), receiver, owner, assets, shares, fee); 
        return shares; 
    }

### DIFF
--- a/contracts/HATVault.sol
+++ b/contracts/HATVault.sol
@@ -576,7 +576,7 @@ contract HATVault is IHATVault, ERC4626Upgradeable, OwnableUpgradeable, Reentran
     function previewWithdrawAndFee(uint256 assets) public view returns (uint256 shares, uint256 fee) {
         uint256 assetsPlusFee = (assets * HUNDRED_PERCENT / (HUNDRED_PERCENT - withdrawalFee));
         fee = assets * withdrawalFee / (HUNDRED_PERCENT - withdrawalFee);
-        shares = _convertToShares(assetsPlusFee, MathUpgradeable.Rounding.Up);
+        shares = _convertToShares(assets + fee, MathUpgradeable.Rounding.Up);
     }
 
     /** @notice See {IHATVault-previewRedeemAndFee}. */
--- a/contracts/HATVault.sol
+++ b/contracts/HATVault.sol
@@ -520,8 +520,7 @@ contract HATVault is IHATVault, ERC4626Upgradeable, OwnableUpgradeable, Reentran
     /** @notice See {IHATVault-withdraw}. */
     function withdraw(uint256 assets, address receiver, address owner) 
         public override(IHATVault, ERC4626Upgradeable) virtual returns (uint256) {
-        uint256 shares = previewWithdraw(assets);
-        uint256 fee = assets * withdrawalFee / (HUNDRED_PERCENT - withdrawalFee);
+        (uint256 shares, uint256 fee) = previewWithdrawAndFee(assets);
         _withdraw(_msgSender(), receiver, owner, assets, shares, fee);
 
         return shares;
@@ -530,8 +529,7 @@ contract HATVault is IHATVault, ERC4626Upgradeable, OwnableUpgradeable, Reentran
     /** @notice See {IHATVault-redeem}. */
     function redeem(uint256 shares, address receiver, address owner) 
         public override(IHATVault, ERC4626Upgradeable) virtual returns (uint256) {
-        uint256 assets = previewRedeem(shares);
-        uint256 fee = _convertToAssets(shares, MathUpgradeable.Rounding.Down) - assets;
+        (uint256 assets, uint256 fee) = previewRedeemAndFee(shares);
         _withdraw(_msgSender(), receiver, owner, assets, shares, fee);
 
         return assets;
@@ -565,16 +563,27 @@ contract HATVault is IHATVault, ERC4626Upgradeable, OwnableUpgradeable, Reentran
     }
 
     /** @notice See {IERC4626Upgradeable-previewWithdraw}. */
-    function previewWithdraw(uint256 assets) public view virtual override(IERC4626Upgradeable, ERC4626Upgradeable) returns (uint256) {
-        uint256 assetsPlusFee = (assets * HUNDRED_PERCENT / (HUNDRED_PERCENT - withdrawalFee));
-        return _convertToShares(assetsPlusFee, MathUpgradeable.Rounding.Up);
+    function previewWithdraw(uint256 assets) public view virtual override(IERC4626Upgradeable, ERC4626Upgradeable) returns (uint256 shares) {
+        (shares,) = previewWithdrawAndFee(assets);
     }
 
     /** @notice See {IERC4626Upgradeable-previewRedeem}. */
-    function previewRedeem(uint256 shares) public view virtual override(IERC4626Upgradeable, ERC4626Upgradeable) returns (uint256) {
-        uint256 assets = _convertToAssets(shares, MathUpgradeable.Rounding.Down);
-        uint256 fee = assets * withdrawalFee / HUNDRED_PERCENT;
-        return assets - fee;
+    function previewRedeem(uint256 shares) public view virtual override(IERC4626Upgradeable, ERC4626Upgradeable) returns (uint256 assets) {
+        (assets,) = previewRedeemAndFee(shares);
+    }
+
+    /** @notice See {IHATVault-previewWithdrawAndFee}. */
+    function previewWithdrawAndFee(uint256 assets) public view returns (uint256 shares, uint256 fee) {
+        uint256 assetsPlusFee = (assets * HUNDRED_PERCENT / (HUNDRED_PERCENT - withdrawalFee));
+        fee = assets * withdrawalFee / (HUNDRED_PERCENT - withdrawalFee);
+        shares = _convertToShares(assetsPlusFee, MathUpgradeable.Rounding.Up);
+    }
+
+    /** @notice See {IHATVault-previewRedeemAndFee}. */
+    function previewRedeemAndFee(uint256 shares) public view returns (uint256 assets, uint256 fee) {
+        uint256 assetsPlusFee = _convertToAssets(shares, MathUpgradeable.Rounding.Down);
+        fee = assetsPlusFee * withdrawalFee / HUNDRED_PERCENT;
+        assets = assetsPlusFee - fee;
     }
 
     /* -------------------------------------------------------------------------------- */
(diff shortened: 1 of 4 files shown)


contracts/HATVault.sol
```diff
@@ -492,13 +492,6 @@ contract HATVault is IHATVault, ERC4626Upgradeable, OwnableUpgradeable, Reentran
     /** @notice See {IHATVault-withdrawRequest}. */
     function withdrawRequest() external nonReentrant {
         IHATVaultsRegistry.GeneralParameters memory generalParameters = registry.getGeneralParameters();
-        // require withdraw to be at least withdrawRequestEnablePeriod+withdrawRequestPendingPeriod
-        // since last withdrawRequest (meaning the last withdraw request had expired)
-        // solhint-disable-next-line not-rely-on-time
-        if (block.timestamp <
```
lirona, line 498: so are we sure we want to remove this? could maybe be nicer for the user to leave it

lirona (commented):
lgtm, i would remove the check `if (shares == 0) revert AmountToDepositIsZero();` from `_deposit` because this is done in `beforeTokenTransfer` now

contracts/HATVault.sol
```diff
@@ -527,18 +520,16 @@ contract HATVault is IHATVault, ERC4626Upgradeable, OwnableUpgradeable, Reentran
...
         return shares;
     }
 
     /** @notice See {IHATVault-redeem}. */
     function redeem(uint256 shares, address receiver, address owner) 
-        public  override(IHATVault, ERC4626Upgradeable) virtual returns (uint256) {
-
+        public override(IHATVault, ERC4626Upgradeable) virtual returns (uint256) {
         uint256 assets = previewRedeem(shares);
         uint256 fee = _convertToAssets(shares, MathUpgradeable.Rounding.Down) - assets;
```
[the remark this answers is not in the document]
ben-kaufman: That's incorrect because assets is already post fee

contracts/HATVault.sol
ben-kaufman: Idk this can be done in the front end so maybe better remove it to save gas

contracts/HATVault.sol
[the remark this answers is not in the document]
lirona: yes so the fee is taken from the post-fee assets, no? what should it be then?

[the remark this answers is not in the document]
ben-kaufman: Well as it is now is fine. But in the suggested change it is taken twice so the calculation will be wrong. Just leaving it as is makes sense I think.

contracts/interfaces/IHATVault.sol
```diff
@@ -178,7 +174,10 @@ interface IHATVault is IERC4626Upgradeable {
     error SystemInEmergencyPause();
     // Cannot set a reward controller that was already used in the past
     error CannotSetToPerviousRewardController();
-
+    // Cannot transfer 0 amount of shares
+    error TransferAmountCannotBeZero();
```
lirona, line 178: maybe change to just `AmountCannotBeZero` since it also includes deposits now

contracts/HATVault.sol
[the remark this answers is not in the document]
lirona: right, so straight forward would be: `_convertToAssets(shares, MathUpgradeable.Rounding.Down) * withdrawalFee / HUNDRED_PERCENT;`
or better to leave as is?

contracts/HATVault.sol
jellegerbrandy: https://github.com/hats-finance/hats-contracts/issues/366

jellegerbrandy: in https://github.com/hats-finance/hats-contracts/issues/366 I argue that we need to keep this check

contracts/HATVault.sol
```diff
@@ -527,18 +520,16 @@ contract HATVault is IHATVault, ERC4626Upgradeable, OwnableUpgradeable, Reentran
     /** @notice See {IHATVault-withdraw}. */
     function withdraw(uint256 assets, address receiver, address owner) 
         public override(IHATVault, ERC4626Upgradeable) virtual returns (uint256) {
-
         uint256 shares = previewWithdraw(assets);
-        uint256 fee = _convertToAssets(shares - _convertToShares(assets, MathUpgradeable.Rounding.Up), MathUpgradeable.Rounding.Up);
+        uint256 fee = assets * withdrawalFee / (HUNDRED_PERCENT - withdrawalFee);
```
jellegerbrandy, line 524: Why is the fee calculation happening here and then passed to _withdraw? Similar logic is duplicated in `redeem`. It all gets confusing bc previewWithdraw takes into account the fees as well.

Perhaps create a `_withdraw(caller, reciever, shares)` returns (assets, fee)` function that handles the calculation in one place?

contracts/HATVault.sol
```diff
@@ -703,6 +693,8 @@ contract HATVault is IHATVault, ERC4626Upgradeable, OwnableUpgradeable, Reentran
         address to,
         uint256 amount
     ) internal virtual override {
+        if (amount == 0) revert TransferAmountCannotBeZero();
+        if (from == to) revert CannotTransferToSelf();
```
jellegerbrandy, line 697: this fixes #369

contracts/HATVault.sol
```diff
@@ -703,6 +693,8 @@ contract HATVault is IHATVault, ERC4626Upgradeable, OwnableUpgradeable, Reentran
         address to,
         uint256 amount
     ) internal virtual override {
+        if (amount == 0) revert TransferAmountCannotBeZero();
```
jellegerbrandy, line 696: this fixes #367

contracts/HATVault.sol
```diff
@@ -760,7 +752,7 @@ contract HATVault is IHATVault, ERC4626Upgradeable, OwnableUpgradeable, Reentran
         // last action was withdrawRequest (and not deposit or withdraw, which
         // reset withdrawRequests[_user] to 0)
         // solhint-disable-next-line not-rely-on-time
-            block.timestamp <=
+            block.timestamp <
```
jellegerbrandy, line 755: fixex #368

contracts/HATVault.sol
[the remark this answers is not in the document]
ben-kaufman: I like it how it is :)

contracts/HATVault.sol
ben-kaufman: Ok I made a function to return both, so fee calculations happen in a single place

contracts/HATVault.sol
```diff
@@ -527,20 +520,16 @@ contract HATVault is IHATVault, ERC4626Upgradeable, OwnableUpgradeable, Reentran
     /** @notice See {IHATVault-withdraw}. */
     function withdraw(uint256 assets, address receiver, address owner) 
         public override(IHATVault, ERC4626Upgradeable) virtual returns (uint256) {
-
-        uint256 shares = previewWithdraw(assets);
-        uint256 fee = _convertToAssets(shares - _convertToShares(assets, MathUpgradeable.Rounding.Up), MathUpgradeable.Rounding.Up);
+        (uint256 shares, uint256 fee) = previewWithdrawAndFee(assets);
```
jellegerbrandy, line 523: fixes #370

contracts/HATVault.sol
```diff
@@ -703,6 +701,8 @@ contract HATVault is IHATVault, ERC4626Upgradeable, OwnableUpgradeable, Reentran
         address to,
         uint256 amount
     ) internal virtual override {
+        if (amount == 0) revert AmountCannotBeZero();
```
jellegerbrandy, line 704: fixes #367

contracts/HATVault.sol
```diff
@@ -492,13 +492,6 @@ contract HATVault is IHATVault, ERC4626Upgradeable, OwnableUpgradeable, Reentran
     /** @notice See {IHATVault-withdrawRequest}. */
     function withdrawRequest() external nonReentrant {
         IHATVaultsRegistry.GeneralParameters memory generalParameters = registry.getGeneralParameters();
-        // require withdraw to be at least withdrawRequestEnablePeriod+withdrawRequestPendingPeriod
```
jellegerbrandy, line 495: fixes #366

jellegerbrandy (approved):

lirona (dismissed):
great

PR #376, opened 2022-09-29
Add arbitrator contract

### LINKED ISSUE #268: create arbitrator contract
The goal is to create a simple arbitrator contract in which the hats governance multisig functions as a simple arbitrator. 

The flow should be is similar to v1, with one crucial difference: in v1,  if governance does not actively approve the claim, the claim will time out and be rejected. In v2, the claim will, by default, be accepted. 

We will write a `HatsGovArbitrator` contract that is owned by the HATs governance multisig and has two functions:
- `approveClaim(vault, claimId)` 
- `dismissClaim(vault, claimId)`

with more or less the following code:

```
function approveClaim(address vault, uint256 claimId) onlyOwner {
       HatVault(vault).challengeClaim(claimId);
       HatVault(vault).approveClaim(claimId);
}

function dismissClaim(address vault, uint256 claimId) onlyOwner {
       HatVault(vault).challengeClaim(claimId);
       HatVault(vault).dismissClaim(claimId);
}
```

ben-kaufman:
Fix #268

### DIFF
--- a/contracts/HATVault.sol
+++ b/contracts/HATVault.sol
@@ -272,7 +272,7 @@ contract HATVault is IHATVault, ERC4626Upgradeable, OwnableUpgradeable, Reentran
                 revert ChallengedClaimCanOnlyBeApprovedByArbitratorUntilChallengeTimeoutPeriod();
             }
             // the arbitrator can update the bounty if needed
-            if(claim.arbitratorCanChangeBounty) {
+            if(claim.arbitratorCanChangeBounty && _bountyPercentage != 0) {
                 claim.bountyPercentage = _bountyPercentage;
             }
         } else {
--- a/contracts/HATVault.sol
+++ b/contracts/HATVault.sol
@@ -272,7 +272,7 @@ contract HATVault is IHATVault, ERC4626Upgradeable, OwnableUpgradeable, Reentran
                 revert ChallengedClaimCanOnlyBeApprovedByArbitratorUntilChallengeTimeoutPeriod();
             }
             // the arbitrator can update the bounty if needed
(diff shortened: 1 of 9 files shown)


contracts/HATVault.sol
```diff
@@ -272,7 +272,7 @@ contract HATVault is IHATVault, ERC4626Upgradeable, OwnableUpgradeable, Reentran
                 revert ChallengedClaimCanOnlyBeApprovedByArbitratorUntilChallengeTimeoutPeriod();
             }
             // the arbitrator can update the bounty if needed
-            if(claim.arbitratorCanChangeBounty) {
+            if(claim.arbitratorCanChangeBounty && _bountyPercentage != 0) {
```
jellegerbrandy, line 275: Are you adding this bc the arbitrator should be be able to call approveClaim but set the bounty to 0 (effectively dismissing the claim). I kind of agree that this is a thing, but this check still allows for a bountypercentage of 0.1%, which is pretty much the same thing. 
I am slightly for not adding this check for that reason

jellegerbrandy (changes requested):
one comment

contracts/HATVault.sol
ben-kaufman: No, the point is to allow the arbitrator to call approve but not change the percentage, it's convenient option since arbitrator could just specify the percentage actually specified, but this just saves mistakes or in the arbitrator contract case, it saves the need to query the contract to know the percentage

jellegerbrandy (dismissed):

lirona (commented):
lgtm

lirona (approved):

PR #379, opened 2022-10-09
Little fixes

lirona:


contracts/interfaces/IHATVaultsRegistry.sol
```diff
@@ -56,15 +56,6 @@ interface IHATVaultsRegistry {
         uint256 claimFee;  
     }
 
-    // How to divide the HAT part of bounties, in percentages (out of {HUNDRED_PERCENT})
```
jellegerbrandy, line 59: good call. Should also remove the reference in the docstring in IHATVAult and in roles.md

jellegerbrandy (changes requested):
1 small change

contracts/interfaces/IHATVaultsRegistry.sol
lirona: right, thanks!

lirona: done

jellegerbrandy (approved):

PR #384, opened 2022-10-25
HATToken to inherit ERC20Votes & ERC20Capped

lirona:


jellegerbrandy (dismissed):
really very nice :)

ben-kaufman:
Looks good, I think we can replace the use of require with custom errors like the rest of the code. Also could make an onlyGovernance modifier instead of the duplicated require

lirona:
> Looks good, I think we can replace the use of require with custom errors like the rest of the code. Also could make an onlyGovernance modifier instead of the duplicated require

sure, now that i got it to work i'll make an interface and make it all pretty :)

contracts/interfaces/IHATToken.sol
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.16;
+
+import "@openzeppelin/contracts/governance/utils/IVotes.sol";
+import "@openzeppelin/contracts/token/ERC20/extensions/draft-IERC20Permit.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
```
ben-kaufman, line 6: Delete this

shayzluf (approved):
@lirona @ben-kaufman approved

PR #401, opened 2022-10-31
New deployment scripts

ben-kaufman:
Replace the old deployment scripts with new ones.


contracts/tokenlock/TokenLockFactory.sol
```diff
@@ -43,10 +43,11 @@ contract TokenLockFactory is CloneFactory, ITokenLockFactory, Ownable {
     /**
      * Constructor.
      * @param _masterCopy Address of the master copy to use to clone proxies
+     * @param _governance Owner of the factory
      */
-    // solhint-disable-next-line func-visibility
-    constructor(address _masterCopy) {
+    constructor(address _masterCopy, address _governance) {
         setMasterCopy(_masterCopy);
+        _transferOwnership(_governance);
```
jellegerbrandy, line 50: not 100% sure this is needed but i guess it does not hurt unless you forget to provide the argument :)

jellegerbrandy (dismissed):

contracts/tokenlock/TokenLockFactory.sol
ben-kaufman: It's because we want the timelockcontroller to be the owner, and like this we can do it in one tx instead of 2

jellegerbrandy (approved):

lirona (approved):
looks good from what i can understand :)

PR #413, opened 2022-11-08
Gas Optmization - submission 15

jellegerbrandy:
Total of averages: 6864750 (-110345 gas compared to original)


contracts/HATTimelockController.sol
```diff
@@ -17,36 +17,36 @@ contract HATTimelockController is TimelockController {
     
     // The following functions are not subject to the timelock
 
-    function approveClaim(HATVault _vault, bytes32 _claimId, uint256 _bountyPercentage) external onlyRole(PROPOSER_ROLE) {
+    function approveClaim(HATVault _vault, bytes32 _claimId, uint256 _bountyPercentage) external payable onlyRole(PROPOSER_ROLE) {
```
lirona, line 20: will save a little gas but we don't want these to be payable

Issue #392, opened 2022-10-29
First depositor can partially steal deposits and DoS vaults


rokinot:
Issue: https://github.com/hats-finance/hats-contracts/issues/392

## Severity
High

## Description

This issue contains code logic from both `HATVault.sol` and `ERC4626Upgradeable.sol` which is an abstract contract inherited by `HATVault.sol`. The first vault's depositor is able to compromise vaults by minting a minimum amount of shares (such as 1 wei of the staking token) and transferring a much larger amount further on (1 ether worth of staking tokens, or 1e18). This makes it so `supply` in `ERC4626Upgradeable.sol` will be 1, while `totalAssets()` will return `1e18`. This makes the share price extremely overvalued, and every future deposit below the transferred amount will revert due to `convertToShares()` returning 0.

Deposit with amounts above this amount transferred by the malicious actor will proceed, however due to rounding issues, the depositor . In the example below, the code has the first malicious actor transferring `1e18` stakingToken amount, a second clueless actor depositing `1.5e18` tokens will only receive 1 share. Since the total token amount is `2.5e18` tokens, 1 share will be able to withdraw `1.25e18` tokens, so the first depositor can exit with a stolen profit of `0.25e18`.

## Mitigation
Uniswap V2 faced the same vulnerability, and fixed it by automatically minting 1000 shares and sending it to address zero. You can check the code snippet [here](https://github.com/Uniswap/v2-core/blob/master/contracts/UniswapV2Pair.sol#L119-L121). For this project, I recommend sending those 1000 shares to address(0) in the `HATVault.sol` initializer.

## Proof of Concept

```
  it("Stealing rewards", async () => {
    await setUpGlobalVars(
      accounts,
      (await web3.eth.getBlock("latest")).number,
      8000,
      [7000, 2500, 500],
      [1000, 500],
      10000,
      0,
      100,
      false,
      2500000,
      60 * 60 * 24 * 3
    );
    currentBlockNumber = (await web3.eth.getBlock("latest")).number;
    var staker = accounts[1];
    var staker2 = accounts[2];
    await stakingToken.approve(vault.address, web3.utils.toWei("1", "ether"), {
      from: staker,
    });
    await stakingToken.approve(vault.address, web3.utils.toWei("1.5", "ether"), {
      from: staker2,
    });
    await stakingToken.mint(staker, "1");

    //Stake minimum amount
    let tx = await vault.deposit("1", staker, { from: staker });

    //Makes sure all values are set properly
[...]

PR #418, opened 2022-11-10
Fix 392

jellegerbrandy:
This PR enforces that the vault always emits minimum of 1M shares (i.e. the minimum deposit is 1USDC$ or 0,000000000001ETH), thereby limiting the rounding errors that can be exploited (cf #392)  to relatively small fractions of the deposits


contracts/HATVault.sol
```diff
@@ -731,6 +734,15 @@ contract HATVault is IHATVault, ERC4626Upgradeable, OwnableUpgradeable, Reentran
             }
         }
     }
+    function _afterTokenTransfer(
+        address from,
+        address to,
+        uint256 amount
+    ) internal virtual override {
+        if (totalSupply() > 0 && totalSupply() < MINIMAL_AMOUNT_OF_SHARES) {
```
ben-kaufman, line 742: Maybe save totalSupply to memory instead of calling storage twice

ben-kaufman (approved):
Looks good, just one small comment

jellegerbrandy:
TODO: make the minimum amoount configurable per vault

PR #420, opened 2022-11-13
call instead of transfer

lirona:


jellegerbrandy (changes requested):
looks good but test coverage needs to be fixed i guess

contracts/HATVaultsRegistry.sol
```diff
@@ -155,8 +155,9 @@ contract HATVaultsRegistry is IHATVaultsRegistry, Ownable {
         if (generalParameters.claimFee > 0) {
             if (msg.value < generalParameters.claimFee)
                 revert NotEnoughFeePaid();
-            // solhint-disable-next-line indent
-            payable(owner()).transfer(msg.value);
+            // solhint-disable-next-line avoid-low-level-calls
+            (bool success,) = payable(owner()).call{value: msg.value}("");
+            require(success, "Failed to send claim fee");
```
ben-kaufman, line 160: Replace require with revert

jellegerbrandy (approved):

PR #422, opened 2022-11-16
417 name internal vars with an underscore

lirona:


hardhat.config.js
```diff
@@ -50,7 +50,7 @@ module.exports = {
     },
     goerli: {
       url: `https://goerli.infura.io/v3/${process.env.INFURA_KEY}`,
-      accounts: [process.env.GOERLI_PK],
+      accounts: process.env.GOERLI_PK,
```
ben-kaufman, line 53: Why was this changed here (also why in other networks no)?

ben-kaufman: Looks like the change breaks CI...

jellegerbrandy: it does fixing it

PR #427, opened 2022-11-20
426 hat token to start untransferable

lirona:


contracts/HATToken.sol
```diff
@@ -29,20 +32,28 @@ contract HATToken is IHATToken, ERC20Votes, ERC20Capped, Ownable {
         emit MinterSet(_minter, _seedAmount);
     }
 
+    function setTransferable() external onlyOwner {
+        transferable = true;
```
jellegerbrandy, line 36: add an event here so the fronted is informed when this happens

test/hattoken.js
```diff
@@ -879,4 +882,56 @@ contract("HATToken", (accounts) => {
...
+
+      try {
+        await token.transfer(
+          accounts[2],
+          web3.utils.toWei("1"),
+          { from: accounts[1] }
+        );
+        assert(false, "cannot transfer before setTransferable");
+      } catch (ex) {
+        assertVMException(ex);
```
jellegerbrandy, line 932:
```suggestion
        assertVMException(ex, "TransfersDisabled");
```
woud this work?

lirona: yes

PR #429, opened 2022-11-22
141-generate-and-publish-docs

lirona:
dodoc extension automatically creates .md files from natspec documentation at every compilation.
these will be published to github pages at:
https://hats-finance.github.io/hats-contracts/
which automatically creates html pages from the `/docs` folder of the `develop` branch


jellegerbrandy (changes requested):
So if at all possible I'd like to avoid committed any generated stuff to our repository. 

It should be possible to generate these docs in a github action, and publish those generated docs in github pages, withouth committing the generated docs to this repo (maybe have the github action update another repo which is then published, idk?)

lirona:
> So if at all possible I'd like to avoid committed any generated stuff to our repository.
> 
> It should be possible to generate these docs in a github action, and publish those generated docs in github pages, withouth committing the generated docs to this repo (maybe have the github action update another repo which is then published, idk?)

what's the problem with it being on the repo? is it ok if it goes to a subdirectory of /docs ? i don't see a simple solution with dodoc but i can try with other tools..

jellegerbrandy (approved):
ok lets see how this plays out :)

PR #459, opened 2022-12-25
Add vault NFT contracts

ben-kaufman:
Moved the vault NFT contracts to the repo here and added it to be part of the deployment script.


jellegerbrandy (approved):
ok just approving this blindly :)

PR #460, opened 2022-12-26
Set swap token

ben-kaufman:
Allow changing the swap token of the HATVaultsRegistry


jellegerbrandy:
closing in favor of #462

PR #461, opened 2022-12-26
Set HATVault implementation

ben-kaufman:
Allow changing the HATVault implementation of the registry


jellegerbrandy:
Closing as we decided not do do this

PR #467, opened 2023-01-11
Upgrade openzeppelin to 4.8.1

jellegerbrandy:


ben-kaufman (approved):

ben-kaufman (changes requested):
4.8.1 was just released with a tiny fix/ change in the ERC4626 so we should use that instead if upgrading.
https://github.com/OpenZeppelin/openzeppelin-contracts/releases/tag/v4.8.1

PR #468, opened 2023-01-11
Hardhat deploy

ben-kaufman:


deploy/001_deploy_hattimelockcontroller.js
```diff
@@ -0,0 +1,46 @@
...
+
+  const { deployer } = await getNamedAccounts();
+  
+  let governance = config["governance"];
+  if (!governance && network.name === "hardhat") {
+    governance = deployer;
+  }
+
+  let hatGovernanceDelay;
+  if (config["timelockDelay"]) {
```
jellegerbrandy, line 18: I think this is not a good pattern, it makes the config files really hard to understand if there are default values with complex logic hidden in deployment scripts.  Better make it obligatory to have all the settings in set explcitly in the config file and fail otherwise (if we must, we can have a `defaults` sectino in the config file as well)

jellegerbrandy:
Perhaps also config.json should become a .js file so we can leave comments

PR #477, opened 2023-03-09
Add 100% max bounty mode

### LINKED ISSUE #480: create the option to liquidate the vault
With the current logic, a vault can not pay out more than 90% in a bounty

We want to add the option to pay out 100% - but this will also permamently end the vault (as all balances will be reset)

* in `submitClaim` and `approveClaim`, allow for the possibility that the bount is 100% (i.e. it must be either <= MAX_BOUNTY_PERCENTAGE, or 100%)
* If the bounty payout is 100%, we "terminate" the vault. This basically means that we should stop deposits to the vault. It seems we get this almost for free from `ERC4626Upgradeable` if we set `_totalSupply` to 0:

```
    function deposit(uint256 assets, address receiver) public virtual override returns (uint256) {
        require(assets <= maxDeposit(receiver), "ERC4626: deposit more than max");
...

    function maxDeposit(address) public view virtual override returns (uint256) {
        return _isVaultCollateralized() ? type(uint256).max : 0;
    }
    function _isVaultCollateralized() private view returns (bool) {
        return totalAssets() > 0 || totalSupply() == 0;
    }
```

ben-kaufman:
Adds the option to set max bounty to 100%. When max bounty is 100%, only payouts of 100% can be made from the vault, and after the payout, depositing to the vault will be disabled forever, so the vault will no longer be usable.
resolves #480

### DIFF
--- a/contracts/HATVault.sol
+++ b/contracts/HATVault.sol
@@ -128,6 +128,7 @@ contract HATVault is IHATVault, ERC4626Upgradeable, OwnableUpgradeable, Reentran
     ArbitratorCanChangeBounty internal arbitratorCanChangeBounty;
 
     bool private _isEmergencyWithdraw;
+    bool private _vaultDestroyed;
 
     modifier onlyRegistryOwner() {
         if (registry.owner() != msg.sender) revert OnlyRegistryOwner();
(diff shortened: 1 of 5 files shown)


contracts/HATVault.sol
```diff
@@ -213,6 +214,8 @@ contract HATVault is IHATVault, ERC4626Upgradeable, OwnableUpgradeable, Reentran
             revert NotSafetyPeriod();
         if (_bountyPercentage > maxBounty)
             revert BountyPercentageHigherThanMaxBounty();
+        if (maxBounty == HUNDRED_PERCENT && _bountyPercentage != HUNDRED_PERCENT)
```
jellegerbrandy, line 217: I am not sure if this is the logic we want. Seems to me a cleaner API is if "maxBounty" continus to mean "the maxium amount you can pay out"  and we enforce a gap here something like
```suggestion
        if (maxBounty == HUNDRED_PERCENT && (_bountyPercentage > MAX_BOUNTY_LIMIT && _bountyPercentage != HUNDRED_PERCENT)
```

jellegerbrandy (changes requested):
looks good, just one small remark

PR #484, opened 2023-03-16
add psibility to have a lower bounty even when maxBounty is 100

jellegerbrandy:


contracts/HATVault.sol
```diff
@@ -214,7 +214,7 @@ contract HATVault is IHATVault, ERC4626Upgradeable, OwnableUpgradeable, Reentran
             revert NotSafetyPeriod();
         if (_bountyPercentage > maxBounty)
             revert BountyPercentageHigherThanMaxBounty();
-        if (maxBounty == HUNDRED_PERCENT && _bountyPercentage != HUNDRED_PERCENT)
+        if (maxBounty == HUNDRED_PERCENT && _bountyPercentage != HUNDRED_PERCENT && _bountyPercentage > MAX_BOUNTY_LIMIT)
             revert PayoutMustBeHundredPercent();
```
ben-kaufman, line 218: Maybe rename the error to `PayoutMustBeUpToMaxBountyLimitOrHundredPercent`
Though that's a bit long :)

PR #501, opened 2023-05-03
Simplify payment splitter address derivation

ben-kaufman:
Make the HATPaymentSplitterFactory derive the address of the splitter contract to be deployed based on the payees and shares used, instead of the deployer's address and nonce.

### DIFF
--- a/contracts/HATPaymentSplitterFactory.sol
+++ b/contracts/HATPaymentSplitterFactory.sol
@@ -8,7 +8,6 @@ import "./HATPaymentSplitter.sol";
 
 contract HATPaymentSplitterFactory {
     address public immutable implementation;
-    mapping(address => uint256) public nonce;
     event HATPaymentSplitterCreated(address indexed _hatPaymentSplitter);
 
     constructor (address _implementation) {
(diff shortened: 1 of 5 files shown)


contracts/HATPaymentSplitterFactory.sol
```diff
@@ -8,24 +8,19 @@ import "./HATPaymentSplitter.sol";
...
-    mapping(address => uint256) public nonce;
     event HATPaymentSplitterCreated(address indexed _hatPaymentSplitter);
 
     constructor (address _implementation) {
         implementation = _implementation;
     }
 
     function createHATPaymentSplitter(address[] memory _payees, uint256[] memory _shares) external returns (address result) {
-        result = Clones.cloneDeterministic(implementation, keccak256(abi.encodePacked(msg.sender, nonce[msg.sender]++)));
+        result = Clones.cloneDeterministic(implementation, keccak256(abi.encodePacked(_payees, _shares)));
```
jellegerbrandy, line 18: but this means that you can only deploy one payment splitter for each _payees,_shares combination. This may not be a problem (it may actually be a feature), but is different from the original implementation (and not just a simplification)

ben-kaufman: Yes but it's intended, maybe should have mentioned that

jellegerbrandy (approved):

PR #506, opened 2023-05-10
Replace revocable enum with bool

ben-kaufman:
This is a small simplification which also saves a bit of contract size. Currently the Revocable option of the token lock is an enum, but only the enabled and disabled options are usable, so this could be simply replaced by a bool instead.


contracts/HATVaultsRegistry.sol
```diff
@@ -341,7 +341,7 @@ contract HATVaultsRegistry is IHATVaultsRegistry, Ownable {
                     generalParameters.hatVestingPeriods,
                     0, // no release start
                     0, // no cliff
-                    ITokenLock.Revocability.Disabled,
+                    false,
```
jellegerbrandy, line 344:
```suggestion
                    false, // not revocable
```

PR #510, opened 2023-05-18
Hackers nft updates

ben-kaufman:
Removes vault, description, and tier from the NFT and use the IPFS hash for the token ID.
Adds option to permanently stop minting of an NFT or multiple NFTs.
Adds HATHackersNFT to the deployement scripts and deploy to Goerli.
Transfer ownership to Hats governance on deployment.


jellegerbrandy (commented):
Sorry for the late review here :)

So I see no bugs here, but the logic of this contract seems very messy to me and I do not understand why


* it seems a bit innatural to have to do two calls (`createNFT` and then `mint` or `mintMultiple`.) Why not do this in a single call (i.e. just have `mint` and that's it). What is the point of mapping ipfshashes in a separate step? 
* `stopMint` does not really "stop" any minting process. What it does is basically say "this tokenId  (or, equilvantly, this ipfsHash) cannot be minted, not now and not at any time in the future".. the implementation seems too complex, bc you already have a kind of whitelist that you manage with "createNFTs"  (i.e the uris mapping), why not use that to "stop" the minting?  But also: since all the functions are `onlyOwner` - what is even the point of having a blacklist of "stopped" unmintable nfts?

ben-kaufman:
Hi, so I'm not sure about the comments.

1. I could add a createAndMint function or something, but I don't think it's in any way unnatural to have the create separated from the minting. The create only happens one, while minting can be called many times. So I think it makes sense to separate them.
2. To use that "whitelist" to stop minting would mean to delete/ alter the uri of the NFT which is not what we want. And the blacklisting to not be able to ever mint (even though it's only owner) was a requirement Shay asked for.

PR #521, opened 2023-08-07
Feat/connetor with hat arbitrator

unknownunknown1:
Arbitrator should call KlerosArbitrator contract once it receives a challenge in order for a dispute to be relayed further to the Court. So `notifyArbitrator()` call should be added inside ` challengeResolution()` function.

Also note that the test file is for the older version of the Connector. It won't pass now but I'll edit it once the contract is established


contracts/HATKlerosV2Connector.sol
```diff
@@ -0,0 +1,140 @@
...
+     *  @param _evidence URI of the evidence to support the challenge.
+     *  Note that the validity of the claim should be checked by Hat arbitrator.
+     */
+    function notifyArbitrator(bytes32 _claimId, string calldata _evidence) external payable {
+        require(msg.sender == address(hatArbitrator), "Wrong caller");
+        require(!claimChallenged[_claimId], "Claim already challenged");
+
+        // Surplus amount will be used to get more votes. Regardless, the UI should demand the correct value.
+        uint256 arbitrationCost = getArbitrationCost();
+        require(msg.value >= arbitrationCost, "Should pay the full deposit.");
```
ben-kaufman, line 88: If msg.vaule is bigger than arbitrationCost the extra funds should be refunded to the caller.

unknownunknown1: Alright, but there is a slight problem that the caller is HatArbitrator in this case, and for the latter to do the refund it should do its own check of `msg.value` and arbitrationCost.

ben-kaufman: Yeah that should be fine

unknownunknown1: I pushed the change

ben-kaufman: Thanks, but this only returns the amount not used, but doesn't move the extra funds back to the hat arbitrator to reimburse the caller no?

unknownunknown1: You're correct, my bad. I pushed the fix

contracts/HATArbitrator.sol
```diff
@@ -532,6 +532,9 @@ contract HATArbitrator {
 
         emit ResolutionChallenged(_vault, _claimId);
 
+        // TODO: add a function to notify the court about the challenge.
```
jellegerbrandy, line 535:
```suggestion
        // TODO: add a function to notify the court about the challenge.
        court.notifyArbitrator(_claimId, _evidence)
```

contracts/HATKlerosConnector.sol
```diff
@@ -0,0 +1,490 @@
...
+    // - 2 side for `Decision.DismissResolution`.
+    struct Round {
+        uint256[3] paidFees; // Tracks the fees paid in this round in the form paidFees[side].
+        bool[3] hasPaid; // True if the fees for this particular side have been fully paid in the form hasPaid[side].
+        mapping(address => uint256[3]) contributions; // Maps contributors to their contributions for each side in the form contributions[address][side].
+        uint256 feeRewards; // Sum of reimbursable appeal fees available to the parties that made contributions to the side that ultimately wins a dispute.
+        uint256[] fundedSides; // Stores the sides that are fully funded.
+    }
+
+    address public immutable governor; // Governor of this contract.
```
jellegerbrandy, line 57: our contracts inherit from Ownable (which also allows for changing ownership), can we also use that here?

contracts/HATKlerosConnector.sol
```diff
@@ -0,0 +1,490 @@
...
+    IArbitrator public immutable klerosArbitrator; // The kleros arbitrator contract (e.g. Kleros Court).
+    HATArbitrator public immutable hatArbitrator; // Address of the Hat arbitrator contract.
+    IHATVault public immutable vault; // Vault contract.
```
jellegerbrandy, line 60: the HATArbitrator contract is written so it can be the arbitrator for different vaults. So this should not be a state variable, but passed as a parameter to the various functions

contracts/HATKlerosConnector.sol
```diff
@@ -0,0 +1,490 @@
...
+        uint256 _loserMultiplier,
+        uint256 _loserAppealPeriodMultiplier
+    ) {
+        emit MetaEvidence(0, _metaEvidence);
+        
+        governor = msg.sender;
+        klerosArbitrator = _klerosArbitrator;
+        arbitratorExtraData = _arbitratorExtraData;
+        hatArbitrator = _hatArbitrator;
+        vault = _vault;
```
jellegerbrandy, line 105: see above

contracts/HATKlerosConnector.sol
```diff
@@ -0,0 +1,490 @@
...
+    // ******************** //
+    
+    /** @dev Notify KlerosArbitrator that expert's committee decision was challenged. Can only be called by Hat arbitrator.
+     *  Requires the arbitration fees to be paid.
+     *  @param _claimId The Id of the active claim in Vault contract.
+     *  @param _evidence URI of the evidence to support the challenge.
+     *  Note that the validity of the claim should be checked by Hat arbitrator.
+     *  @return surplusFee The amount of fee that exceeds required arbitration cost.
+     */
+    function notifyArbitrator(bytes32 _claimId, string calldata _evidence) external payable returns (uint256 surplusFee) {
```
jellegerbrandy, line 155: needs a `_vault` parameter as well

contracts/HATArbitrator.sol
jellegerbrandy: @unknownunknown1 could you fill in these calls, as I cannot write to your PR

contracts/HATKlerosConnector.sol
```diff
@@ -0,0 +1,490 @@
...
+        require(msg.sender == address(hatArbitrator), "Wrong caller");
+        require(!claimChallenged[_claimId], "Claim already challenged");
+
+        uint256 arbitrationCost = getArbitrationCost();
+        require(msg.value >= arbitrationCost, "Should pay the full deposit.");
```
jellegerbrandy, line 160: the hats disputes are being pid for by erc20 tokens. Do we have to use ETH?

jellegerbrandy: in that way, we can also use transferFrom and do not need to reimburse logic

contracts/HATKlerosConnector.sol
```diff
@@ -0,0 +1,490 @@
...
+
+        DisputeStruct storage dispute = disputes.push();
+        dispute.claimId = _claimId;
+
+        // Preemptively create a new funding round for future appeals.
+        dispute.rounds.push();
+
+        uint256 externalDisputeId = klerosArbitrator.createDispute{value: arbitrationCost}(RULING_OPTIONS,  arbitratorExtraData);
+        dispute.externalDisputeId = externalDisputeId;
+        externalIDtoLocalID[externalDisputeId] = localDisputeId;
```
jellegerbrandy, line 175: I feel like I aksed this before, but can you remind me why we cannot use the externalDisputeId as a local reference as well?

contracts/HATKlerosConnector.sol
```diff
@@ -0,0 +1,490 @@
...
+            payable(msg.sender).send(surplusFee);
+        }
+
+        emit Challenged(_claimId);
+        emit Dispute(klerosArbitrator, externalDisputeId, metaEvidenceUpdates, localDisputeId);
+        emit Evidence(klerosArbitrator, localDisputeId, msg.sender, _evidence);
+    } 
+
+    /** @dev Give a ruling for a dispute. Can only be called by the arbitrator.
+     *  @param _disputeId ID of the dispute in the arbitrator contract.
```
jellegerbrandy, line 187:
```suggestion
     *  @param _disputeId ID of the dispute in the Kleros arbitrator contract.
```

contracts/HATKlerosConnector.sol
```diff
@@ -0,0 +1,490 @@
...
+
+        uint256 localDisputeId = disputes.length;
+
+        DisputeStruct storage dispute = disputes.push();
+        dispute.claimId = _claimId;
+
+        // Preemptively create a new funding round for future appeals.
+        dispute.rounds.push();
+
+        uint256 externalDisputeId = klerosArbitrator.createDispute{value: arbitrationCost}(RULING_OPTIONS,  arbitratorExtraData);
```
jellegerbrandy, line 173:
```suggestion
        uint256 klerosDisputeId = klerosArbitrator.createDispute{value: arbitrationCost}(RULING_OPTIONS,  arbitratorExtraData);
```

contracts/HATKlerosV2Connector.sol
```diff
@@ -0,0 +1,141 @@
...
+    struct DisputeStruct {
+        bytes32 claimId; // Id of the claim in HATVault contract.
+        uint256 externalDisputeId; // Id of the dispute created in Kleros court.
+        Decision ruling; // Ruling given by the arbitrator.
+        bool resolved; // True if the dispute has been resolved.
+    }
+
+    IArbitrator public immutable klerosArbitrator; // The kleros arbitrator contract (e.g. Kleros Court).
+    HATArbitrator public immutable hatArbitrator; // Address of the Hat arbitrator contract.
+    IHATVault public immutable vault; // Vault contract.
```
ben-kaufman, line 41: The contract should be used for all vaults, not just for a single vault. I see this is only used in the rule function, can we make the rule function accept a _vault parameter? If not we'll need to save the vault for a specific dispute ID at an earlier stage, maybe when notifyArbitrator is called.

contracts/HATKlerosConnector.sol
```diff
@@ -0,0 +1,490 @@
...
+
+        uint256 multiplier;
+        {
+            uint256 winner = klerosArbitrator.currentRuling(externalDisputeId);
+            if (winner == _side) {
+                multiplier = winnerMultiplier;
+            } else {
+                require(
+                    block.timestamp - appealPeriodStart <
+                        (appealPeriodEnd - appealPeriodStart).mulCap(loserAppealPeriodMultiplier) / MULTIPLIER_DIVISOR,
```
jellegerbrandy, line 256: as I understand it, you are comparing here an absolute timestamp with an interval, and so I would guess this checkw ould almost always fail. Or do I misunderstand?

contracts/HATKlerosConnector.sol
```diff
@@ -0,0 +1,486 @@
...
+    DisputeStruct[] public disputes; // Stores the disputes created in this contract.
+    mapping(bytes32 => bool) public claimChallenged; // True if the claim was challenged in this contract..
+    mapping(uint256 => uint256) public override externalIDtoLocalID; // Maps external dispute ids to local dispute ids.
+
+    /** @dev Raised when a claim is challenged.
+     *  @param _claimId Id of the claim in Vault cotract.
+     */
+    event Challenged(bytes32 indexed _claimId);
+    
+    modifier onlyGovernor {require(msg.sender == governor, "The caller must be the governor."); _;}
```
ben-kaufman, line 77: I think it will be better to use Ownable here instead of this. We might want to allow transferring ownership.

contracts/HATKlerosConnector.sol
```diff
@@ -0,0 +1,486 @@
...
+        // Preemptively create a new funding round for future appeals.
+        dispute.rounds.push();
+
+        uint256 externalDisputeId = klerosArbitrator.createDispute{value: arbitrationCost}(RULING_OPTIONS,  arbitratorExtraData);
+        dispute.externalDisputeId = externalDisputeId;
+        externalIDtoLocalID[externalDisputeId] = localDisputeId;
+
+        emit Challenged(_claimId);
+        emit Dispute(klerosArbitrator, externalDisputeId, metaEvidenceUpdates, localDisputeId);
+        emit Evidence(klerosArbitrator, localDisputeId, msg.sender, _evidence);
```
ben-kaufman, line 183: msg.sender here is the arbitrator contract, but I think it should be the actual disputer. So you might want to ask for the disputer address as a parameter to the function

contracts/HATKlerosConnector.sol
```diff
@@ -0,0 +1,486 @@
...
+    // ******************** //
+    
+    /** @dev Notify KlerosArbitrator that expert's committee decision was challenged. Can only be called by Hat arbitrator.
+     *  Requires the arbitration fees to be paid.
+     *  @param _claimId The Id of the active claim in Vault contract.
+     *  @param _evidence URI of the evidence to support the challenge.
+     *  Note that the validity of the claim should be checked by Hat arbitrator.
+     *  @return surplusFee The amount of fee that exceeds required arbitration cost.
+     */
+    function notifyArbitrator(bytes32 _claimId, string calldata _evidence) external payable returns (uint256 surplusFee) {
```
ben-kaufman, line 155: It would be useful to have an interface for V1 and V2 connectors which has this function so we can use it in the arbitrator contract

contracts/HATKlerosV2Connector.sol
unknownunknown1: I see. Rule() function is a part of IArbitrable interface so we can't change it, but we can store it when arbitrator is notified, indeed. We can tie vault address to a dispute ID and then retrieve this address in rule() function

contracts/HATKlerosConnector.sol
```diff
@@ -0,0 +1,486 @@
...
+        if (round.fundedSides.length == 1) finalRuling = round.fundedSides[0];
+
+        dispute.ruling = Decision(finalRuling);
+        dispute.resolved = true;
+
+        bytes32 claimId = dispute.claimId;
+        if (finalRuling == uint256(Decision.DismissResolution)) {            
+            hatArbitrator.dismissResolution(vault, claimId); //
+        } else {
+            // Arbitrator sided with committee or refused to arbitrate (gave 0 ruling).
```
ben-kaufman, line 211: I'm not sure if we want to dismiss or execute the resolution if arbitrator doesn't give a decision. @jellegerbrandy ?

contracts/HATKlerosConnector.sol
```diff
@@ -0,0 +1,486 @@
...
+        }
+      
+        emit Ruling(IArbitrator(msg.sender), _disputeId, finalRuling);
+    }
+
+    /** @dev Submit a reference to evidence. EVENT.
+     *  @param _localDisputeId The id of the related dispute.
+     *  @param _evidenceURI Link to evidence.
+     */
+    function submitEvidence(uint256 _localDisputeId, string calldata _evidenceURI) external override {
```
ben-kaufman, line 222: I think you should also check the dispute actually exists, maybe by checking claimId is not 0

contracts/HATKlerosConnector.sol
```diff
@@ -0,0 +1,490 @@
...
+    uint256 public metaEvidenceUpdates; // Relevant index of the metaevidence.
+    bytes public arbitratorExtraData; // Extra data for the arbitrator.
+
+    uint256 public winnerMultiplier; // Multiplier for calculating the appeal fee that must be paid for the ruling that was chosen by the arbitrator in the previous round, in basis points.
+    uint256 public loserMultiplier; // Multiplier for calculating the appeal fee that must be paid for the ruling that the arbitrator didn't rule for in the previous round, in basis points.
+    uint256 public loserAppealPeriodMultiplier; // Multiplier for calculating the duration of the appeal period for the loser, in basis points.
+
+    DisputeStruct[] public disputes; // Stores the disputes created in this contract.
+    mapping(bytes32 => bool) public claimChallenged; // True if the claim was challenged in this contract..
+    mapping(uint256 => uint256) public override externalIDtoLocalID; // Maps external dispute ids to local dispute ids.
```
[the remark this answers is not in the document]
unknownunknown1: This name is a part of IDisputeResolver interface. Can't change it

contracts/HATKlerosConnector.sol
```diff
@@ -0,0 +1,490 @@
...
+    }
+
+    // ******************** //
+    // *    Challenges    * //
+    // ******************** //
+    
+    /** @dev Notify KlerosArbitrator that expert's committee decision was challenged. Can only be called by Hat arbitrator.
+     *  Requires the arbitration fees to be paid.
+     *  @param _claimId The Id of the active claim in Vault contract.
+     *  @param _evidence URI of the evidence to support the challenge.
```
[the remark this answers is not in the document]
unknownunknown1: Should be doable, however I'm not very knowledgeable on evidence handling. If we encounter issues here I'll consult with the team

contracts/HATKlerosConnector.sol
unknownunknown1: I'll add it

contracts/HATKlerosConnector.sol
unknownunknown1: Sure

contracts/HATKlerosConnector.sol
unknownunknown1: It's not an absolute timestamp but the difference between the current time and the time when the dispute started. We check whether we passed half of the appeal timeout or not

contracts/HATKlerosV2Connector.sol
```diff
@@ -0,0 +1,141 @@
...
+        ExecuteResolution, // Accept the claim as it is without changing.
+        DismissResolution // Dismiss the claim.
+    }
+
+    struct DisputeStruct {
+        bytes32 claimId; // Id of the claim in HATVault contract.
+        uint256 externalDisputeId; // Id of the dispute created in Kleros court.
+        Decision ruling; // Ruling given by the arbitrator.
+        bool resolved; // True if the dispute has been resolved.
+    }
```
ben-kaufman, line 37: You could move this to a shared interface/ base contract with the V1 connector to avoid duplication

contracts/HATKlerosConnector.sol
jellegerbrandy: if you mean the check on line 216 `!dispute.resolved`, this will pass also for non-existing disputes, so that will not work

contracts/HATKlerosConnector.sol
```diff
@@ -0,0 +1,481 @@
...
+        require(!dispute.resolved, "Dispute already resolved");
+
+        emit Evidence(klerosArbitrator, _localDisputeId, msg.sender, _evidenceURI);
+    }
+
+    // ************************ //
+    // *       Appeals        * //
+    // ************************ //
+
+    /** @dev Takes up to the total amount required to fund a side. Reimburses the rest. Creates an appeal if both sides are fully funded.
```
jellegerbrandy, line 225:
```suggestion
    /** @dev Takes up to the total amount required to fund a side. Reimburses the rest. Creates an appeal if two of the three sides are fully funded.
```

contracts/HATKlerosConnector.sol
```diff
@@ -0,0 +1,481 @@
...
+     */
+    function changeLoserMultiplier(uint256 _loserMultiplier) external onlyOwner {
+        loserMultiplier = _loserMultiplier;
+    }
+
+    /** @dev Changes loserAppealPeriodMultiplier variable.
+     *  @param _loserAppealPeriodMultiplier The new loserAppealPeriodMultiplier value.
+     */
+    function changeLoserAppealPeriodMultiplier(uint256 _loserAppealPeriodMultiplier) external onlyOwner {
+        loserAppealPeriodMultiplier = _loserAppealPeriodMultiplier;
```
jellegerbrandy, line 120: should be <= than MULTIPLIER_DIVISOR

contracts/HATKlerosConnector.sol
```diff
@@ -0,0 +1,481 @@
...
+        uint256 _loserAppealPeriodMultiplier
+    ) {
+        emit MetaEvidence(0, _metaEvidence);
+        
+        klerosArbitrator = _klerosArbitrator;
+        arbitratorExtraData = _arbitratorExtraData;
+        hatArbitrator = _hatArbitrator;
+        winnerMultiplier = _winnerMultiplier;
+        loserMultiplier = _loserMultiplier;
+        loserAppealPeriodMultiplier = _loserAppealPeriodMultiplier;
```
jellegerbrandy, line 95: should be <= than MULTIPLIER_DIVISOR

contracts/HATArbitrator.sol
```diff
@@ -532,6 +535,8 @@ contract HATArbitrator {
 
         emit ResolutionChallenged(_vault, _claimId);
 
+        IHATKlerosConnector(court).notifyArbitrator(_claimId, _evidence, _vault, msg.sender);
```
ben-kaufman, line 538: Also need to send the ETH here in the call (and make the function payable)

contracts/HATKlerosConnector.sol
ben-kaufman: Sorry right, thought for a moment it was a mapping

contracts/HATKlerosConnector.sol
ben-kaufman: Shouldn't it only be possible to fund the yes or no options? Then have the neutral option be counted as a no?

contracts/HATKlerosConnector.sol
```diff
@@ -0,0 +1,481 @@
...
+        bytes32 claimId; // Id of the claim in HATVault contract.
+        uint256 externalDisputeId; // Id of the dispute created in Kleros court.
+        Decision ruling; // Ruling given by the arbitrator.
+        bool resolved; // True if the dispute has been resolved.
+        IHATVault vault; // Address of the vault related to a dispute.
+        Round[] rounds; // Appeal rounds.
+    }
+
+    // Round struct stores the contributions made to particular sides.
+    // - 0 side for `Decision.None`.
```
ben-kaufman, line 41: Do we want to allow people to actively support the None option? Wouldn't it make sense to just have people support yes or no, then if it's None by the arbitrator just treat it as a no?

jellegerbrandy: apparently this is baked in in V2 (right @unknownunknown1 ?) so we just copy that logic here

jellegerbrandy (approved):

PR #542, opened 2023-11-22
Create HATAirdrop

ben-kaufman:


contracts/tge/HATAirdrop.sol
```diff
@@ -0,0 +1,111 @@
...
+    error LeafAlreadyRedeemed();
+    error InvalidMerkleProof();
+    error CannotRecoverBeforeDeadline();
+
+    using SafeERC20Upgradeable for IERC20Upgradeable;
+
+    bytes32 public root;
+    uint256 public startTime;
+    uint256 public deadline;
+    uint256 public lockEndTime;
```
shayzluf, line 27: we should consider lock length

contracts/tge/HATAirdrop.sol
```diff
@@ -0,0 +1,111 @@
...
+    uint256 public periods;
```
shayzluf, line 28: this can be calculated param(once a day)

ben-kaufman: But to calculate this would mean we need to hardcode a period length (or pass it and then it's just the same no?)

ben-kaufman: So I think it's more flexible just directly passing periods count

contracts/tge/HATAirdrop.sol
ben-kaufman: So here we can either just start from a certain date to a certain date, or we can start on claim and end after X time. I implemented the former but we the latter is also a valid option, we just need to decide what behavior we want here.

contracts/tge/HATAirdrop.sol
```diff
@@ -0,0 +1,101 @@
...
+    mapping (bytes32 => bool) public leafRedeemed;
+
+    event MerkleTreeSet(string _merkleTreeIPFSRef, bytes32 _root, uint256 _startTime, uint256 _deadline);
+    event TokensRedeemed(address indexed _account, address indexed _tokenLock, uint256 _amount);
+
+    constructor () {
+        _disableInitializers();
+    }
+
+    function initialize(
```
jellegerbrandy, line 53: this function needs a docstring

jellegerbrandy: for example, it is not clear that the `startTime` the start both of the airdrop and the vesting schedule..

contracts/tge/HATAirdrop.sol
```diff
@@ -0,0 +1,101 @@
...
+        string memory _merkleTreeIPFSRef,
```
jellegerbrandy, line 54: do we need this on-chain? Perhaps add in the doc then that this is supposed to be the merkle tree itself (is it?)

contracts/tge/HATAirdropFactory.sol
```diff
@@ -0,0 +1,115 @@
...
+        uint256 _totalAmount,
+        IERC20Upgradeable _token,
+        ITokenLockFactory _tokenLockFactory
+    );
+
+    constructor (address _implementation) {
+        implementation = _implementation;
+    }
+
+    function updateImplementation(address _newImplementation) external onlyOwner {
```
jellegerbrandy: this is nice but it is not as flexible as it looks, bc any implementation contract needs to use the same signature for the initialization function. If we want the flexibility to create new airdrop contracts, we proably should pass the arguments encoded in a single parameter (but also that has it's drawbacks in terms of ux)

contracts/tge/HATAirdrop.sol
```diff
@@ -0,0 +1,101 @@
...
+        if (block.timestamp < startTime) revert CannotRedeemBeforeStartTime();
+        // solhint-disable-next-line not-rely-on-time
+        if (block.timestamp > deadline) revert CannotRedeemAfterDeadline();
+        bytes32 leaf = _leaf(_account, _amount);
+        if (leafRedeemed[leaf]) revert LeafAlreadyRedeemed();
+        if(!_verify(_proof, leaf)) revert InvalidMerkleProof();
+        leafRedeemed[leaf] = true;
+
+        address _tokenLock = address(0);
+        if (lockEndTime != 0) {
```
jellegerbrandy:
```suggestion
        if (lockEndTime > block.timestamp) {
```

contracts/tge/HATAirdropFactory.sol
```diff
@@ -0,0 +1,115 @@
...
+
+    function createHATAirdrop(
+        string memory _merkleTreeIPFSRef,
+        bytes32 _root,
+        uint256 _startTime,
+        uint256 _deadline,
+        uint256 _lockEndTime,
+        uint256 _periods,
+        uint256 _totalAmount,
+        IERC20Upgradeable _token,
```
jellegerbrandy: do we want to airdrop tokens different than hats? IF not, perhaps this argument should be in the constructor?

contracts/tge/HATAirdropFactory.sol
```diff
@@ -0,0 +1,115 @@
...
+    constructor (address _implementation) {
+        implementation = _implementation;
+    }
+
+    function updateImplementation(address _newImplementation) external onlyOwner {
+        implementation = _newImplementation;
+        emit ImplementationUpdated(_newImplementation);
+    }
+
+    function withdrawTokens(IERC20Upgradeable _token, uint256 _amount) external onlyOwner {
```
jellegerbrandy: So this allows for rugpulling the airdrop, as we discussed, and to mitigate we expect the funtion to be timelocked. Do we want a comment explaining that?

contracts/tge/HATAirdropFactory.sol
```diff
@@ -0,0 +1,115 @@
...
+        address owner = owner();
```
jellegerbrandy: probably save a miniscule amount of gas 
```suggestion
        address owner = msg.sender;
```

contracts/tge/HATAirdropFactory.sol
jellegerbrandy: If users do not redeem their tokens from the airdrop contract bf the deadline, hats can get those tokens back calling withdrawTokens. However, unclaimed tokens in the tokenlock will be lost forever. This can kind of be grief-attacked by an attacker that calls `redeem` on all unclaimed leafs just before the deadline (or even frontrunning `withdrawTokens`)

contracts/tge/HATAirdrop.sol
jellegerbrandy: I don't get what you are saying Ben. We know the duration (lockEndTime - startTime). If we want to hardcode "once a day" as shay suggests, we can calculate periods as `(lockEndTime - startTime)/24h`

jellegerbrandy (commented):
- the pattern "redeem -> tokenlock.release()` to get your tokens is pretty bad ux: it is hard to understand, and esp for small amounts it is a lot of gas costs. One limit case is where all tokens are already vested on the `redeem()` call - in that case it would be much nicer (chepaer, easier) to just get the tokens directly (see my suggestion there). 

- it would be cool to have some way to claim several airdrops in a single tx. Something like `factory.redeem(airDropContracts[], proofs[], amounts[], account)`

contracts/tge/HATAirdrop.sol
ben-kaufman: Yes that's what I'm saying that we'd need to hardcode it, but why would we want to hardcode a value we are not sure about? It's better to leave periods completely flexible by letting the deployer pass it

contracts/tge/HATAirdrop.sol
ben-kaufman: It's just emitted in the event not saved on chain, just the IPFS url

contracts/tge/HATAirdropFactory.sol
ben-kaufman: Ok yeah it'd be more flexible

contracts/tge/HATAirdropFactory.sol
ben-kaufman: I'm not sure if we might do different tokens or not

contracts/tge/HATAirdropFactory.sol
ben-kaufman: So you'd recommend allowing only beneficiary to redeem for themselves right?

contracts/tge/HATAirdrop.sol
```diff
@@ -0,0 +1,118 @@
...
+        _disableInitializers();
+    }
+
+    /**
+    * @notice Initialize a HATAirdrop instance
+    * @param _merkleTreeIPFSRef new merkle tree ipfs reference.
+    * @param _root new merkle tree root to use for verifying airdrop data.
+    * @param _startTime start of the redeem period and of the token lock (if exists)
+    * @param _deadline end time to redeem from the contract
+    * @param _lockEndTime end time for the token lock contract (if exists)
```
jellegerbrandy:
```suggestion
    * @param _lockEndTime end time for the token lock contract. If this date is in the past, the tokens will be transferred directly to the user and no token lock will be created
```

jellegerbrandy (dismissed):
lgtm, just one small update on the docstring

jellegerbrandy (approved):

PR #587, opened 2024-11-24
Add ExpertCommitteeArbitrator

ben-kaufman:
Add ExpertCommitteeArbitrator


contracts/interfaces/IHATClaimsManager.sol
shayzluf, line 1: is it possible to update this interface without redeploying the registry @ben-kaufman ?

shayzluf (commented):
could those changes be applied with the current registry?

jellegerbrandy (dismissed):
it looks good to me

